PR #5, opened 2016-03-21
Remove trailing slash in blog "read more" links

takidau:
Trailing slash appears to work fine in my local version of Jekyll, but the live site chokes on it. Removing it is technically more correct anyway, I believe.


evilsoapbox:
LGTM!

PR #6, opened 2016-03-23
Fix 'getting started page' link in homepage alert message

mrtns:


davorbonaci:
Thanks Martin.

LGTM. We'll merge shortly.

@evilsoapbox et al., do you know is it perhaps possible to automate hyperlink validation with the Jekyll build process? This sounds like something that could be possible.

evilsoapbox:
@davorbonaci Good question (sorry for delay - NEXT). I created JIRA https://issues.apache.org/jira/browse/BEAM-150 to handle this. I will either find the method to do it or hammer our a Ruby script to do it and hook it into the jekyll create process.

PR #7, opened 2016-03-29
Addition of contribution guide

james-malone:


content/README.md
```diff
@@ -0,0 +1,73 @@
+# Apache Beam (incubating) website
```
davorbonaci: A little puzzled why this file showed up in the `content` directory. Accidental / intentional?

davorbonaci:
LGTM.

content/README.md
evilsoapbox: Just pushed a very minor tweak to _config.yml to ignore the README (and Ruby bits)

PR #9, opened 2016-04-02
Capability matrix updates for sliding window support in Spark Runner

takidau:
- Add sliding window support for batch Spark runner.
- Center capability matrix (expand/collapse looks cleaner).
- Default to summary view (easier to grok).
- Add "Last updated" timestamp.

### DIFF
--- a/_data/capability-matrix.yml
+++ b/_data/capability-matrix.yml
@@ -249,8 +249,8 @@ categories:
             l2: supported
             l3: ''
           - class: spark
-            l1: 'No'
-            l2: ''
+            l1: 'Partially'
+            l2: currently only supported in batch
             l3: ''
 
       - name: Session windows
--- a/_pages/capability-matrix.md
+++ b/_pages/capability-matrix.md
@@ -4,7 +4,9 @@ title: "Apache Beam Capability Matrix"
 permalink: /capability-matrix/
 ---
 
+
 # Apache Beam Capability Matrix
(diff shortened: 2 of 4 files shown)


amitsela:
+1 on sliding windows support for batch in the Spark runner

evilsoapbox:
LGTM overall - my only question, should we explicitly define which versions of all of these we're using for the comparison? Do we always implicitly use the latest stable release, something else?

davorbonaci:
LGTM.

(Good question regarding versioning. I don't think we need to worry about this before the first stable release of Beam. Before that, all information should refer to the most-recent state of the project. Once we release a stable version, perhaps an annotation would be appropriate. E.g., "available since version 1.2.3.")

evilsoapbox:
@davorbonaci Sorry, should have been more specific. What about the other projects? We may want to mention specific (**stable**) releases for each runner so people see a Beam->Runner version compatibility matrix as well.

davorbonaci:
Good point.

(I guess you probably meant Beam Runner -> Backend matrix, as opposed to Beam -> Runner, and to mention specific releases of each backend, right?)

Definitely a nice-to-have. (Not sure how easy is it to gather that information, however.)

takidau:
Created [BEAM-166](https://issues.apache.org/jira/browse/BEAM-166) for version tracking in the matrix. Thanks!

PR #19, opened 2016-05-24
[BEAM-103] update capability matrix

mxm:
This reflects the changes of BEAM-103 in the Capability Matrix.


mxm:
CC @evilsoapbox @davorbonaci

PR #20, opened 2016-06-01
Add Flink Batch Runner Blog Post

aljoscha:
I'd like to add this short post about the recent changes to the Flink Batch runner. The date is set in the future, we'll probably decide when to schedule this if we merge it.

R: @davorbonaci 
R: @evilsoapbox 
R: @mxm


francesperry:
Nice!

davorbonaci:
Looks great to me!

Personally, I'd probably be a little more conservative and avoid claiming "100%" or "very confident in correct results for all pipelines". There are bugs everywhere, particularly those that occur at high scale. Perhaps you could give yourself a way out by using a term that still implies *all*, but isn't as strong as *100%*.

aljoscha:
@davorbonaci yeah, you're right, that's a bit too bold. I'm relaxing it.

evilsoapbox:
Looks great! I am happy to merge later in June to correspond with the publish date if you'd like.

aljoscha:
Thanks @evilsoapbox. The date is just a place holder, though. You can publish whenever you see fit. I somehow thought that you're the person who has a bit of an eye on the blog posts, so that we space them out properly.

PR #23, opened 2016-06-20
HTML/link cleanup + test(s)!

james-malone:
This PR does a few things:
- Adds support for testing the built website via the `rake` command using the `html-proofer` module
  - Checks HTML to ensure it is valid
  - Looks for bad links
  - Ignores the `javadocs` because they are a complete mess
- Cleanup for the website based on an initial test of the built content

This is a first step in getting the site built by Travis CI (future PR will add travis config.)


evilsoapbox:
Note - will squash the merges once this is LGTM.

content/Rakefile
```diff
@@ -0,0 +1,10 @@
+require 'html-proofer'
```
davorbonaci: Probably shouldn't be here.

Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
```
davorbonaci: Do these files have to be in the root directory? Is it possible to move them elsewhere?

content/beam/capability/2016/03/17/capability-matrix.html
```diff
@@ -154,19 +154,23 @@ <h1 class="post-title" itemprop="name headline">Clarifying & Formalizing Runner
     
     
 
-    <td width="25%" class="cap-summary" style="background-color:#ec3;border-color:#ca1"></td>
+    <td width="25%" class="cap-summary" style="background-color:#ec3;border-color:#ca1"><center><b>&#x2713;</b></center>
```
davorbonaci: I think there's a distinct possibility that we are using different versions -- and we are stomping on each others generated code, and producing a lot of updates in each PR (unnecessarily). Not sure can we do something about it -- seems unlikely.

Gemfile
evilsoapbox: @davorbonaci A Rubyist may be more opinionated, but AFAIK these are always *supposed* to be in the root directory of the app per the [spec](http://bundler.io/man/gemfile.5.html)

content/beam/capability/2016/03/17/capability-matrix.html
evilsoapbox: @davorbonaci That looks like one of the explicit "fixes" to create valid HTML. We were embedding a `<content>` tag inside a `<b>` tag which is apparently invalid HTML. I swapped them in the following files which I suspect might be the reason for that update:

- `_includes/capability-matrix-row-blog.md`
- `_includes/capability-matrix-row-full.md`
- `_includes/capability-matrix-row-summary.md`

davorbonaci:
LGTM. Please self-merge.

evilsoapbox:
ack!

PR #24, opened 2016-06-22
Travis CI YAML config

james-malone:
Adding a Travis CI configuration file so we can use Travis on the Beam website.

PR #28, opened 2016-07-25
Adding Travis CI YAML config

james-malone:
Re-adding YAML config for Travis CI (sorry - had to fix a GitHub mess I accidentally created here.) This replaces PR #24.

### DIFF
--- a/content/capability-matrix/index.html
+++ b/content/capability-matrix/index.html
@@ -99,7 +99,7 @@
 
       <div class="container">
         <h1 id="apache-beam-capability-matrix">Apache Beam Capability Matrix</h1>
-<p><span style="font-size:11px;float:none">Last updated: 2016-07-12 08:21 PDT</span></p>
+<p><span style="font-size:11px;float:none">Last updated: 2016-07-27 13:19 PDT</span></p>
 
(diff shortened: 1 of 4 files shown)


dhalperi:
Glad the PR found that JB's website is down.. shows the checks are working.

Is there a way to whitelist that particular failure?

evilsoapbox:
You mean avoid URL status checking entirely or to have it be non-fatal?

dhalperi:
Latter

evilsoapbox:
Does not look like it. While that may make some tests flaky, I also posit that it means our live content may have HTTP 4/X 5/X errors, which is probably worth knowing about with a fatal.

dhalperi:
What about, e.g., https://github.com/gjtorikian/html-proofer#ignoring-content ?

evilsoapbox:
We can do that, though we'll possibly have to ignore links on a per-link basis or in general.

How about we ignore this one link for the time being and if this becomes a problem, we can mod Jekyll to inject the ignore attribute for all links in one area (like the Blog) or in all areas?

dhalperi:
ack

evilsoapbox:
New commit (550be51) with minor update to ignore link added. If it looks good, I can squash them and merge.

PR #29, opened 2016-08-02
[BEAM-510] releases: use ASF download script to balance load on Apache

dhalperi:
From: http://mail-archives.apache.org/mod_mbox/incubator-general/201608.mbox/%3C7E0226B1-0386-499C-8473-61A8E51A691B%40classsoftware.com%3E

> All links to the mirrored distribution artifacts must not reference the
> main Apache web site. They should use the standard mechanisms to
> distribute the load between the mirrors.

http://www.apache.org/dev/release-download-pages.html#download-scripts

### DIFF
--- a/use/releases.md
+++ b/use/releases.md
@@ -8,7 +8,7 @@ redirect_from: /releases/
 # Apache Beam Releases
 
 The easiest way to use Apache Beam is via one of the released versions in the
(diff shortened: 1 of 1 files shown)


evilsoapbox:
LGTM - noticed we switch from https to http. I assume that is also intentional to lower overhead.

dhalperi:
Fixed to HTTPS. not strictly required by Apache, but good policy.

Also rebased on top of @francesperry 's recent refactoring. Will merge.

PR #35, opened 2016-08-04
Add a link to the Apache Code of Conduct

dhalperi:
R: @francesperry 

I saw that Apache Fluo (incubating) had this link, and I liked it. https://www.apache.org/foundation/policies/conduct#specific-guidelines

I did not see that this was required by the Podling rules, but IMO there's no harm in being explicit that we want to create a good environment. This is Apache policy anyway, I don't think Beam would be exempt if we wanted to be.

[@jbonofre -- does this need discussion on the Beam list?]


evilsoapbox:
My two cents - good idea to include the link.

jbonofre:
Good idea. No need to be discussed on the mailing list IMHO.

LGTM

francesperry:
Agreed. Good idea. LGTM

dhalperi:
Thanks all. Self-merged.

PR #36, opened 2016-08-08
Revise Beam programming guide for new DoFn

kennknowles:
R: @francesperry

### DIFF
--- a/content/learn/programming-guide/index.html
+++ b/content/learn/programming-guide/index.html
@@ -435,10 +435,10 @@
 </code></pre>
 </div>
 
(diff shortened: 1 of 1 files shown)


kennknowles:
Actual changes are in the first commit, rebuild in the second.

learn/programming-guide.md
```diff
@@ -271,11 +271,11 @@ A `DoFn` processes one element at a time from the input `PCollection`. When you
 static class ComputeWordLengthFn extends DoFn<String, Integer> { ... }
 ```
 
-Inside your `DoFn` subclass, you'll need to override the method `processElement`, where you provide the actual processing logic. You don't need to manually extract the elements from the input collection; the Beam SDKs handle that for you. Your override of `processElement` should accept an object of type `ProcessContext`. The `ProcessContext` object gives you access to an input element and a method for emitting an output element:
+Inside your `DoFn` subclass, you'll write a method annotated with `@ProcessElement` where you provide the actual processing logic. You don't need to manually extract the elements from the input collection; the Beam SDKs handle that for you. Your `@ProcessElement` method should accept an object of type `ProcessContext`. The `ProcessContext` object gives you access to an input element and a method for emitting an output element:
```
francesperry: I'm not sure what Devin's plans were for making this language-independent. If this is general prose, we should avoid java-specific implementation details. But that's probably a topic for another day ;-)

francesperry:
LGTM. Will merge.

PR #39, opened 2016-09-05
[BEAM-102] update capability matrix

mxm:
This updates the matrix to the most recent development status.

CC @aljoscha

### DIFF
--- a/_data/capability-matrix.yml
+++ b/_data/capability-matrix.yml
@@ -121,7 +121,7 @@ categories:
             l3: Batch implemented supports a distributed implementation, but streaming mode may force some size restrictions. Neither mode is able to push lookups directly up into key-based sources.
           - class: flink
             l1: 'Yes'
-            l2: fully supported
+            l2: some size restrictions in streaming
             l3: Batch implemented supports a distributed implementation, but streaming mode may force some size restrictions. Neither mode is able to push lookups directly up into key-based sources.
           - class: spark
             l1: 'Partially'


aljoscha:
I think it would be good to use more or less the description that we have for the Dataflow runner. This pretty accurately describes the features/limitations.

mxm:
> Batch implemented supports a distributed implementation, but streaming mode may force some size restrictions. Neither mode is able to push lookups directly up into key-based sources. 

Will add this.

_data/capability-matrix.yml
```diff
@@ -120,10 +120,9 @@ categories:
             l2: some size restrictions in streaming
             l3: Batch implemented supports a distributed implementation, but streaming mode may force some size restrictions. Neither mode is able to push lookups directly up into key-based sources.
           - class: flink
-            jira: BEAM-102
-            l1: 'Partially'
-            l2: no supported in streaming
-            l3: Supported in batch. Side inputs for streaming are currently WiP.
+            l1: 'Yes'
+            l2: fully supported
```
aljoscha: could be "some size restrictions in streaming" as for Dataflow, just nitpicking ... 😅 🙊

mxm: That's fine. Thanks!

aljoscha:
LGTM now!

mxm:
Actually there is a typo in the comment above. Will fix before merging.

PR #45, opened 2016-10-14
Fix jekyll setup instructions

cschneider:
When I just tried the jekyll setup on a recent ubuntu the start of jekyll failed with an error.
It reported that it needs an activesupport < 5.0.0 while the version the listed commands installed was 5.0.1.

What I found is the workaround below. I guess this needs to be fixed in the jekyll gems too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Before working with the Jekyll code, you will need to install Jekyll:
     $ gem install jekyll
     $ gem install jekyll-redirect-from
     $ gem install html-proofer
+    $ gem install activesupport -v 4.2.7
 
 *If you are on a Mac, you may need to install
 [Ruby Gems](https://rubygems.org/pages/download).*


jbonofre:
R: @jbonofre
CC: @francesperry

README.md
```diff
@@ -27,6 +27,7 @@ Before working with the Jekyll code, you will need to install Jekyll:
     $ gem install jekyll
     $ gem install jekyll-redirect-from
     $ gem install html-proofer
+    $ gem install activesupport -v 4.2.7
```
jbonofre: I didn't need this on Ubuntu.

What's your system ?

More over, it's a bit risky to use a specific gem version.

francesperry: I don't believe I've needed this on two recent fresh installs on Mac, though I did need 'bundler', which was just added in pull/42. (Also I recall general frustration with 'gem' installation not seeming to do the right transitive things.)

I know @tgroh, @melap, and @eljefe6a have recently gone through this -- any thoughts from you folks?

eljefe6a: I checked back through my notes and didn't need that on the Mac.

melap: Yes, I had the same problem @cschneider had (on Ubuntu) about 2 weeks ago, and had to install the older version of activesupport.

francesperry: BTW, while investigating syntax highlighting dependencies, I've just discovered how the Gemfile works: http://jekyll.tips/jekyll-casts/gemfiles-and-the-bundler/

Ours is currently forcing a specific version of activesupport for reasons I don't understand. Will dig in more tonight and let you know what I find.

francesperry: See https://github.com/apache/incubator-beam-site/pull/50 

Hopefully that solves the issue so this PR is unnecessary?

jbonofre: +1 no need with the change in #50

cschneider: @francesperry Sounds good. The specific version I used was only a workaround anyway

francesperry:
@cschneider Can you close this PR, since it's obsolete? Thanks!

PR #50, opened 2016-10-21
Turn on syntax highlighting & fix gem installation

francesperry:
- Updated README to correctly use Bundler to manage gem dependencies.
- Turned on syntax highlighting by using an existing style sheet.

R: @davorbonaci 
CC: @cschneider, @jbonofre


francesperry:
Need to dig in and see why tests are failing (pass locally). Will do that tomorrow.

jbonofre:
It sounds good, let me test deeper. Thanks !

francesperry:
* Got tests passing after many extraneous pushes to trigger Travis. 
* Added use of --deployment to the instructions.

Also CC: @evilsoapbox, who has ruby experience.

PR #51, opened 2016-10-21
[BEAM-501] New skin proposal

jbonofre:
This a new skin proposal. It's still a work in progress, but it already allows people to take a look, provide feedback, and upgrade.


aljoscha:
Would it be possible to host a preview somewhere?

jbonofre:
@aljoscha yup, it's already available there: http://maven.nanthrax.net/beam/ (just the homepage)

davorbonaci:
I'm on this! Review & merge coming up shortly!

davorbonaci:
Merged (site not rebuilt).

PR #59, opened 2016-11-02
[BEAM-752] Infrastructure for toggling code snippets in documentation.

abdullah353:
## Working

This commit will enable mark down language to support two languages `java` and `python`, By adding a nav bar on top of code blocks.

Add html class as `language-python` or `language-java` and the HTML elements visibility will be controlled based on user selected language preferences.

Code blocks in markdown languages starting with *```java* or *```python* have by default `langauge-java` or `language-python` classes so no special attributes required to apply this effect.

## Example

1) Toggles for code block.

Add python code below the java code in markdown language, It will be converted into selectable code block div.

> `` `java "some-code-in-java"

> `` `python "sam-code-in-python"

And user will see block like this.

screen shot 2016-11-02 at 9 47 30 am

2) Section of text.

Wrap a text with attributes having class `language-java` or `language-python` and it's visibility will be changed based on user preferences.

> To use Beam, your driver program must first create an instance of the Beam SDK class `Pipeline` <span class="language-java">(typically in the `main()` function)</span><span class="language-python">(typically in `def`)</span>.

When `java` is selected then this will render as:
screen shot 2016-11-02 at 9 50 54 am

And for `python` it will render as:

screen shot 2016-11-02 at 9 51 23 am

3) User preference persistence:

The language preference will be stored on HTML5 local storage, keeping it persistence per user bases.

### DIFF
--- a/js/language-switch.js
+++ b/js/language-switch.js
@@ -3,6 +3,8 @@ $(document).ready(function(){
     var elAttr = "div[class^=\"language-\"]";
     // Default language.
     var defLang = "language-java";
+    // Wrapper class for css stylings.
+    var wrapper="language-switcher"
 
     /**
       * @desc Generate bootstrapped like nav template, showing supported languages in nav.
@@ -16,7 +18,7 @@ $(document).ready(function(){
             langList+="<li data-lang=\""+languages[i]+"\"> \
                 <a>"+languages[i].replace("language-","")+"</a></li>";
         }
-        return "<div class=\"language-switcher "+languages.join(" ")+"\"> \
+        return "<div class=\""+wrapper+" "+languages.join(" ")+"\"> \
                 <ul class=\"nav nav-tabs\">"+langList+"</ul> </div>";
     }
 
@@ -27,7 +29,7 @@ $(document).ready(function(){
     */
     function getLang(text) {
         var lang = /language-\w+/i.exec(text)
-        return (lang)? lang[0]:'';
+        return (lang)? lang[0]:"";
     }
 
     /**
@@ -43,7 +45,7 @@ $(document).ready(function(){
             return lang;
         }
 
-        lang.push(getLang(el.attr('class')))
+        lang.push(getLang(el.attr("class")))
         return nextFetch(el.next(), lang)
     }
 
@@ -51,17 +53,12 @@ $(document).ready(function(){
       * @desc Perform language switch on page, and persist it's state.
     */
     var switchLanguage = function() {
-        var langPref = localStorage.getItem('langPref') || defLang;
-
-        // If selected language has no code, then revert back to default.
-        if($("div."+langPref).length === 0){
-            langPref = defLang;
-        }
+        var langPref = localStorage.getItem("codePreference") || defLang;
 
         // Adjusting active elements in navigation header.
-        $(".language-switcher li").removeClass('active').each(function(){
+        $("."+wrapper+" li").removeClass("active").each(function(){
             if($(this).data("lang") === langPref) {
-                $(this).addClass('active');
+                $(this).addClass("active");
             }
         });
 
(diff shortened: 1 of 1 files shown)


evilsoapbox:
Just did a pull and test. Overall looks good to me. I did notice one quirk which I think is unavoidable. If there are no Python code blocks (programming guide as an example) then when you tab to Python, you don't see anything. I added one test block and it worked; the rest of the tabs for Python on the page showed blank content, which seems OK.

evilsoapbox:
First of all, thanks for doing this! This is a really good start. After an in-depth test of your fork, I noticed a few quirks and have a few ideas:

1. Somehow syntax hi-lighting seems to have been lost in the code blocks now. This is a breaking bug.
2. If a block has no code you see a tab but nothing below it - is there any way we can render a small div or block at least for visual consistency (or better yet, hide the tab entirely)?

I'd also love to use this all over the site, for walkthroughs, docs, etc. This will be *super* useful!

francesperry:
To extend James comment, it would be great if we could also use this to toggle runner-specific things like execution instructions with Flink, Dataflow, Spark, Gearpump, Apex tabs!

Very excited to see this ;-)

davorbonaci:
If I may follow along as well -- very happy to see this. Thanks @mabdullah353, much appreciated.

mabdullah353:
Thank you very much guys for reviewing and giving feedback 👍 👍 

@evilsoapbox  
## Response to 1 point: (striked):

Code highlight shouldn't be disturbed by this script. But if previously it was showing then i love to fix it, Is their any task in Asana you familiar regarding this issue please feel free to assign it to me :)

## Response to 2 point:

Hiding is better option, i fixed it now.

## Response to 3 point:

I have added auto code detection on pages, i think it's better than hardcoding languages, Also based on @francesperry suggestion (which i liked very much) making dynamic at this stage is good.

_**Current Woking as:**_
If markdown has code block as:

> ''' java

> ''' python

> ''' html

Then it will be rendered as:
screen shot 2016-11-03 at 4 43 17 pm

Similarly:
> ''' java

Will result into a single tab.
screen shot 2016-11-03 at 5 13 52 pm

CSS Improvement
==========

I removed the bootstrap classes from script. Now tabs are shown in `left` by default, because of default bootstrap styling. If we want to change the positioning we can add this simple style in css.

```css
.language-switcher .nav-tabs > li {
    float: right;
}
```
And it will result as:
screen shot 2016-11-03 at 5 41 01 pm

Change log:
==========

* Added auto detection of languages, by looking all codes block available on page.
* Fixed empty navigation tabs will not be visible, when their is no code to show against the selected language.
* Nav will only show the languages in which code is available.
* Removed bootstrap grid classes from script. To give control to add custom styling from css.

mabdullah353:
@francesperry  looking forward to integrate in other repos as well. What's best way to start this ? By searching a relative task in respective repos and add pull request over their ?

I have written python version of code against the available `programming-guide`, But i am not able to find task in JIRA , Am i allowed to `create` task in JIRA, unable to find any guide in `Apache Beam Contribution Guide` any thoughts ?

davorbonaci:
LGTM. I'm going to merge, since we have a few people who want to use it right away!

Also, I'll add a few comments shortly as ideas for improvement.

davorbonaci:
To answer your JIRA question -- please use it without worry. File new issues, comment on existing ones, resolve completed ones, or do any other action you find appropriate.

I'll try to clarify @francesperry comment: Beam is a set of SDKs and a set of runners. This pull request creates a toggle for SDK languages, which is fantastic. We could use the same thing for runners too. For example, on the Quickstart page, we should have a command to run a pipeline. That command will be slightly different based on the runner; having a toggle there too would be great! (In that case, "Direct" runner would be the default runner, similarly as Java is the default SDK.)

Another idea would be to extend the toggle for larger sections, not just code blocks -- such that you can mix&match text and code within the content of the toggle tab.

In summary, great work! We'd love to get you involved even more -- please reach out anytime if one of us can be of any assistance.

francesperry:
Yup -- what @davorbonaci said.  I would love to have Direct/Spark/Flink/Dataflow/etc toggles in the quickstart I'm working on!

PR #60, opened 2016-11-03
add a symlink to current Javadoc

dhalperi:
R: @francesperry @davorbonaci @evilsoapbox 
CC: @melap per request

Note this is a filesystem-level symlink; in practice it appears that Jekyll will generate two copies of the content rather than a HTTP 301/302 to redirect. I figured symlink is easier to maintain than a custom 302 page, so this is the lesser evil.

But there may be a better way to do this.

### DIFF
--- a/learn/sdks/javadoc/current
+++ b/learn/sdks/javadoc/current
@@ -1,0 +1,1 @@
+0.3.0-incubating


dhalperi:
(Jekyll cannot support 302, that's a webserver thing. Of course. https://github.com/jekyll/jekyll-redirect-from/issues/49)

davorbonaci:
Does [`redirect_from` and `redirect_to`](https://github.com/jekyll/jekyll-redirect-from) work? It should for the top-level page, but probably won't for deeper direct links. We currently use it elsewhere to avoid breaking old links.

dhalperi:
No, `redirect_to` is actually worse.

`javadoc/current` -> `javadoc/0.3.0-incubating`.

But `javadoc/current/test.txt` -> redirect loop on `javadoc/current/0.3.0-incubating`.

dhalperi:
So I think the question is whether we want `current` in URLs or whether all links on this site should simply use `{{ var.latest_release }}` in place of `0.3.0-incubating`.

I learn towards the former (`current` actually in URLs, but the content changes), because the variable-solution only works when used on this website -- but not for links pasted in other places.

davorbonaci:
Symlinks seem to break on [other platforms](http://stackoverflow.com/questions/5917249/git-symlinks-in-windows), without an elaborate setup. So, I'd discourage that.

I'd suggest a path of least resistance -- a Jekyll global variable. It works for what we need it, plus a few other nice benefits, without sacrificing portability.

dhalperi:
Abandoned

PR #65, opened 2016-11-04
[BEAM-890] Update compatibility matrix for Spark.

amitsela:


amitsela:
R: @jbonofre

davorbonaci:
R: @kennknowles

_data/capability-matrix.yml
```diff
@@ -50,8 +50,8 @@ categories:
             l3: "Uses Flink's keyBy for key grouping. When grouping by window in streaming (creating the panes) the Flink runner uses the Beam code. This guarantees support for all windowing and triggering mechanisms."
           - class: spark
             l1: 'Partially'
-            l2: group by window in batch only
-            l3: "Uses Spark's groupByKey for grouping. Grouping by window is currently only supported in batch."
+            l2: support for grouping by panes (streaming) is a work in progress.
+            l3: Using groupByKey for grouping, but only if the pipeline explicitly calls for GroupByKey or the model forces it. For efficient group-compute see Combine.
```
kennknowles: Say something positive about batch first, like:

"Full support for in batch mode. GroupByKey with multiple trigger firings in streaming mode is a work in progress."

And then I don't think you actually need to teach users about how to program against it here. The statement is likely true for most runners. If you just say "Using Spark's \`groupByKey\`" that tells users what they need to know, if they are familiar with Spark. (do use code font, I'd say)

_data/capability-matrix.yml
```diff
@@ -88,7 +88,7 @@ categories:
           - class: spark
             l1: 'Yes'
             l2: fully supported
-            l3: Supports GroupedValues, Globally and PerKey.
+            l3: Using combineByKey and aggregate functions.
```
kennknowles: "Using Spark's \`combineByKey\` and \`aggregate\` functions."

_data/capability-matrix.yml
```diff
@@ -124,9 +124,9 @@ categories:
             l2: some size restrictions in streaming
             l3: Batch mode supports a distributed implementation, but streaming mode may force some size restrictions. Neither mode is able to push lookups directly up into key-based sources.
           - class: spark
-            l1: 'Partially'
-            l2: not supported in streaming
-            l3: "Side input is actually a broadcast variable in Spark so it can't be updated during the life of a job. Spark-runner implementation of side input is more of an immutable, static, side input."
+            l1: 'Yes'
+            l2: fully supported
+            l3: A side input is actually a broadcast variable in Spark. In streaming mode, a side input could be updated between micro-batches. The distribution of side inputs to workers is not partitioned, but only a worker assigned with a relevant task will get a copy of the side input.
```
kennknowles: "Using Spark's broadcast variables."

I actually think the rest of it is just a description of what a side input is. They are always global views of a PCollection, and are generally always only read by workers who are working on a task that needs them.

I'm not saying you can't have some explanation, but maybe more focused. Maybe the second sentence is good, like "In streaming mode, side input values only update between micro-batches."

_data/capability-matrix.yml
jbonofre: Agree with Amit.

amitsela: @kennknowles you mean use the `tt` tag for `groupByKey`, `combineByKey` and such ?

_data/capability-matrix.yml
amitsela: I don't understand the here for `Flink` and `Dataflow` - clearly side inputs have size restrictions by design.. why is streaming different then batch ? and why write it here ?

_data/capability-matrix.yml
kennknowles: Whatever Jekyll requires. I guess it is YAML and I was thinking Markdown.

amitsela: @jbonofre any input here ? I'm the worst web-developer in the project. For sure.

jbonofre: I think ' should do the trick.

amitsela:
@kennknowles @jbonofre trying a 2nd iteration.

davorbonaci:
LGTM

jbonofre:
LGTM

PR #66, opened 2016-11-04
Test baseurl presence and fix breaks

dhalperi:
Two commits:

1. Add baseurl everywhere it's needed
2. Add tests that will catch future breaks

The tests use a separate config to set test-specific properties. The `site.url` configuration specifically does not appear settable from the Jekyll CLI. Unfortunately this means that there is one property shared between config and test (`.testcontent`)

Also adds `content` and `.testcontent` to the excludes list so that the builds of each do not include the other.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,7 +1,9 @@
+require 'fileutils'
 require 'html-proofer'
 
 task :test do
-  sh "bundle exec jekyll build"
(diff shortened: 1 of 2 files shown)


dhalperi:
R: @davorbonaci

davorbonaci:
LGTM. Merged.

PR #70, opened 2016-11-07
[BEAM-902] Add runner toggles

abdullah353:
Hi,
You can use runner in the following way.

> {:.runner-<name-of-your-runner-1>}
> '''
> your runner 1 code.

> {:.runner-<name-of-your-runner-2>}
> '''
> your runner 2 code.

And it will get converted to tabs, just like for programming languages.

I have updated the {src/get-started/quickstart.md} MarkDown as well.

screen shot 2016-11-07 at 11 49 20 am

I have created class from the script, so that the same code can be used for runner toggling as well.

Kindly review, it.
Thanks.

@francesperry  @davorbonaci


davorbonaci:
Beautiful. LGTM. Merging.

PR #77, opened 2016-11-11
[BEAM-508] Fill in the documentation/runners/dataflow portion of the website

melap:
R: @davorbonaci @francesperry

PR #76, opened 2016-11-11
[BEAM-505] Fill in the documentation/runners/direct portion of the website

melap:
R: @davorbonaci @francesperry

PR #79, opened 2016-11-12
[BEAM-899] Add Flink Instructions to quickstart.md

aljoscha:
I added a section about runner specific dependencies in the Quickstart.

R: @francesperry

### DIFF
--- a/src/get-started/quickstart.md
+++ b/src/get-started/quickstart.md
@@ -7,7 +7,7 @@ redirect_from:
   - /getting-started/
 ---
 
-# Apache Beam Java SDK Quickstart 
+# Apache Beam Java SDK Quickstart
 
 This Quickstart will walk you through executing your first Beam pipeline to run [WordCount]({{ site.baseurl }}/get-started/wordcount-example), written using Beam's [Java SDK]({{ site.baseurl }}/documentation/sdks/java), on a [runner]({{ site.baseurl }}/documentation#runners) of your choice.
 
@@ -16,7 +16,7 @@ This Quickstart will walk you through executing your first Beam pipeline to run 
 
 
 ## Set up your Development Environment
- 
+
 1. Download and install the [Java Development Kit (JDK)](http://www.oracle.com/technetwork/java/javase/downloads/index.html) version 1.7 or later. Verify that the [JAVA_HOME](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/envvars001.html) environment variable is set and points to your JDK installation.
 
 1. Download and install [Apache Maven](http://maven.apache.org/download.cgi) by following Maven's [installation guide](http://maven.apache.org/install.html) for your specific operating system.
@@ -24,7 +24,7 @@ This Quickstart will walk you through executing your first Beam pipeline to run 
 
 ## Get the WordCount Code
 
-The easiest way to get a copy of the WordCount pipeline is to use the following command to generate a simple Maven project that contains Beam's WordCount examples and builds against the most recent Beam release: 
+The easiest way to get a copy of the WordCount pipeline is to use the following command to generate a simple Maven project that contains Beam's WordCount examples and builds against the most recent Beam release:
 
 ```
 $ mvn archetype:generate \
@@ -38,7 +38,7 @@ $ mvn archetype:generate \
       -Dpackage=org.apache.beam.examples
 ```
 
-This will create a directory `word-count-beam` that contains a simple `pom.xml` and a series of example pipelines that count words in text files. 
+This will create a directory `word-count-beam` that contains a simple `pom.xml` and a series of example pipelines that count words in text files.
 
 ```
 $ cd beam-word-count/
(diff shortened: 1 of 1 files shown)


src/get-started/quickstart.md
```diff
@@ -53,6 +53,20 @@ MinimalWordCount.java	WordCount.java
 
 For a detailed introduction to the Beam concepts used in these examples, see the [WordCount Example Walkthrough]({{ site.baseurl }}/get-started/wordcount-example). Here, we'll just focus on executing `WordCount.java`.
 
+## Runner-Specific Dependencies
```
davorbonaci: I don't think this should be needed. The quickstart is using examples/WordCount archetype, which already has (or, should have) this set up correctly. So, no action needed from the user.

I think this generally falls under `Ensure you've done any runner-specific setup` bullet below -- we should put it on the runner page. See #76 and #77 for good examples, IMO.

src/get-started/quickstart.md
```diff
@@ -74,8 +88,9 @@ $ mvn compile exec:java -Dexec.mainClass=org.apache.beam.examples.WordCount \
 ```
 
 {:.runner-flink}
-``` 
-TODO BEAM-899
+```
+$ mvn compile exec:java -Dexec.mainClass=org.apache.beam.examples.WordCount \
```
davorbonaci: Could we teach the users here how to run on Flink locally and on a remote cluster?

src/get-started/quickstart.md
```diff
@@ -101,8 +116,8 @@ $ ls counts*
 ```
 
 {:.runner-flink}
-``` 
-TODO BEAM-899
+```
+$ ls counts*
```
davorbonaci: Is there some UI that users can see if running locally/remotely?

content/get-started/quickstart/index.html
```diff
@@ -147,6 +147,7 @@ <h1 id="apache-beam-java-sdk-quickstart">Apache Beam Java SDK Quickstart</h1>
 <ul id="markdown-toc">
   <li><a href="#set-up-your-development-environment" id="markdown-toc-set-up-your-development-environment">Set up your Development Environment</a></li>
   <li><a href="#get-the-wordcount-code" id="markdown-toc-get-the-wordcount-code">Get the WordCount Code</a></li>
+  <li><a href="#runner-specific-dependencies" id="markdown-toc-runner-specific-dependencies">Runner-Specific Dependencies</a></li>
```
davorbonaci: FYI only -- we are trying the practice of avoiding `content/*` in the pull requests, and generate new content when merging PR. Much easier to handle. (Of course, works either way.)

src/get-started/quickstart.md
aljoscha: As of now, if you follow the Quickstart steps you will end up with a pom that only has the direct runner and dataflow runner as dependencies. (This is tracked in https://issues.apache.org/jira/browse/BEAM-905) I think we either have to make BEAM-905 a blocker for this or add in specific instructions.

src/get-started/quickstart.md
aljoscha: I was thinking of putting that in the Flink Runner-specific page to not overload this because it requires some more involved steps.

If you want, I can also put that here, though.

src/get-started/quickstart.md
aljoscha: See comment above.

src/get-started/quickstart.md
davorbonaci: This is already done in `examples`; I think I already saw a PR for archetypes too. So, I'd go with the assumption this will be done very shortly. Sure, we'd need a release, but there are other rough edges too -- perhaps we can point to nightlies at the top of the page.

src/get-started/quickstart.md
davorbonaci: Wouldn't this be a simple pipeline option with hostname/port only to point to an already configured cluster? (How to set up such a cluster would be overload for sure.)

src/get-started/quickstart.md
davorbonaci: Please do as you see fit.

From my perspective, I'd point the users to it, but not teach them how to set up a cluster.

src/get-started/quickstart.md
aljoscha: Unfortunately not, the Flink Runner does not automatically pick up class files and jars like the Dataflow runner does. You need to create a jar that contains all the dependencies (which a `mvn package` does because we have maven shade in the examples archetype) and then run it as you would run a Flink job, i.e. something like `bin/flink run -c my.WordCountClass my-jarfile.jar -my-parameters`.

Also, there is a bug right now which requires users to place the jar into the lib folder of the Flink installation. Something with Flink not using the proper Classloader in a certain place.

I can still put that in, though. What do you think?

davorbonaci: Hm. Perhaps a separate tab for the time being? `Flink-local` and `Flink-remote`?

davorbonaci: I think so. Javascript doesn't hardcode anything.

aljoscha: I'm sorry, what do you mean by that?

Also, should we maybe put all of the runner specific instructions into tabs that are labeled with the runner? As it is now, we just have four consecutive boxes of shell commands that are not labeled. I imagine the confusion to only get bigger once we add more runners to this.

aljoscha:
@davorbonaci I updated this to have tabs for `Flink-local` and `Flink-cluster`. Also, I'm now using the `-Pflink-runner` profile which doesn't exist yet in the archetype but which is also used for the apex instructions. This will only work once we update the archetypes.

davorbonaci:
LGTM. A clear improvement, so merging. Thanks @aljoscha.

* Test failure is due to a broken link to the Apex feature branch. It's being fixed elsewhere.
* Instead of `--runner=org.apache.beam.runners.flink.FlinkRunner`, `--runner=flink` should be enough, no?

aljoscha:
For running it locally yes, for the `bin/flink` invocation no, because the class path is (probably) not setup correctly and it doesn't seem to pick up the runners so these short names don't work.

PR #80, opened 2016-11-13
Add Jenkins job definitions for existing projects using Jenkins DSL

davorbonaci:
R: @dhalperi @jasonkuster 

It works beautifully for the "test" project. It doesn't work for the "stage" project due to jenkinsci/ghprb-plugin#417.

### DIFF
--- a/tools/common_job_properties.groovy
+++ b/tools/common_job_properties.groovy
@@ -1,0 +1,106 @@
+// Contains functions that help build Jenkins projects. Functions typically set
+// common properties that are shared among all Jenkins projects.
+class common_job_properties {
+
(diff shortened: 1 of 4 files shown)


davorbonaci:
@dhalperi, @jasonkuster, I found a way to solve the problem. I think this should be ready to go. (Of course, the final test can only be done after merging.) So happy to have this completed!

davorbonaci:
After 39 attempts, it is all working now!

jasonkuster:
LGTM. This is going to be an awesome improvement.

davorbonaci:
Thanks! Merging.

PR #82, opened 2016-11-14
"before" instead of "behind" for "early" elements

manuzhang:
R: @tgroh

### DIFF
--- a/src/_posts/2016-10-20-test-stream.md
+++ b/src/_posts/2016-10-20-test-stream.md
@@ -78,7 +78,7 @@ reactions to speculative and late panes and dropped data.
 
 ## Element Timings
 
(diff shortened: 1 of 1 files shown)


tgroh:
This LGTM, plus reasoning (mainly written for my sake)

We're answering one of two questions (from [Streaming 102](https://www.oreilly.com/ideas/the-world-beyond-batch-streaming-102)): "Where in event time are results calculated?" or "When in processing time are results materialized?"

This describes the arrival of elements with regard to the watermark, which is part of when we materialize results. Given that, it is correct to use temporal rather than spatial words.

manuzhang:
would be great if we can allow comments on blog

PR #86, opened 2016-11-15
Adjust main page header to look like v2's, only with pulldowns

francesperry:
Starting the process of merging in the new v2/ landing page by porting the header to the new style.

Still todo: center beam image, center menus, add top orange bar.

Reviewer: @davorbonaci


francesperry:
(This path is unlikely to work -- css merging nightmares.)

PR #90, opened 2016-11-21
[BEAM-1007] Runner Toggles in Quickstart.md Don't Work in Safari

abdullah353:
@aljoscha This will fix your issue.


aljoscha:
Thanks a lot for the quick fix! 😃 

I'm merging...

PR #91, opened 2016-11-21
[BEAM-277] Add transforms section to the programming guide

melap:
Adds the remaining content for the transforms section: Combine, Flatten, Partition, Side inputs and side outputs


melap:
R: @davorbonaci

davorbonaci:
This looks good to me. Merging.

There seem to be issues with rendering both Java and Python SDK links, and not toggling them. However, that doesn't seem to be related to this pull request. So, merging.

PR #92, opened 2016-11-21
Update download page to specify backward-compatibility guarantees

davorbonaci:
R: @francesperry or @bchambers or @jbonofre 

This was previously discussed on the mailing list, but we should probably refresh this discussion.


francesperry:
LGTM -- matches previous discussions on dev@

davorbonaci:
Thanks! Merged.

PR #99, opened 2016-12-01
Simplify Flink Runner instructions for running on cluster

aljoscha:
R: @davorbonaci I think this should be simple/straighforward enough now 😃

### DIFF
--- a/src/get-started/quickstart.md
+++ b/src/get-started/quickstart.md
@@ -88,12 +88,9 @@ $ mvn compile exec:java -Dexec.mainClass=org.apache.beam.examples.WordCount \
 
 {:.runner-flink-cluster}
 ```
-$ mvn package -Pflink-runner
-$ cp target/word-count-beam-bundled-0.1.jar /path/to/flink/lib/
-$ bin/flink run -c org.apache.beam.examples.WordCount lib/word-count-beam-0.1.jar  \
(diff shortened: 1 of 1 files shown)


src/get-started/quickstart.md
```diff
@@ -88,12 +88,9 @@ $ mvn compile exec:java -Dexec.mainClass=org.apache.beam.examples.WordCount \
 
 {:.runner-flink-cluster}
 ```
-$ mvn package -Pflink-runner
-$ cp target/word-count-beam-bundled-0.1.jar /path/to/flink/lib/
-$ bin/flink run -c org.apache.beam.examples.WordCount lib/word-count-beam-0.1.jar  \
-    --inputFile=/path/to/quickstart/pom.xml  \
-    --output=/tmp/counts \
-    --runner=org.apache.beam.runners.flink.FlinkRunner
+$ mvn package exec:java -Dexec.mainClass=org.apache.beam.examples.WordCount \
```
davorbonaci: /path/to/quickstart/pom.xml --> pom.xml?
(for consistency)

(the reference target/ is already dependent on the current directory)

aljoscha: I made it like this because the jar path can be relative to the current directory while `inputFile` and `output` file usually are paths to somewhere in a distributed filesystem. What do you think?

PR #104, opened 2016-12-09
[BEAM-438] Rename PTransform apply to expand throughout

kennknowles:
The term does not appear to have been used in the running text, so replacement in the code of the gaming examples is the only change outside of the Javadoc.

This PR should perhaps wait until `0.4.0-incubating` is released, depending on what we point users to these days.

R: @davorbonaci


davorbonaci:
LGTM. Merging.

(May cause some confusion temporarily, but still good to go.)

PR #109, opened 2016-12-15
Publish 0.4.0-incubating release

jbonofre:


jbonofre:
R: @davorbonaci

kennknowles:
So the second Javadoc commit is actually a second try as part of 0.4.0-incubating, right?

I would prefer the "Publish 0.4.0-incubating release" commit to not touch generated files. Since the Javadoc is overwritten anyhow, can you remove it from this commit?

Something like this (I haven't tested it)

```
git rebase -i
# edit that commit
git checkout -- documentation/sdks/javadoc
git commit --amend
git rebase --continue
```

jbonofre:
The second commit is updated based on the RC3 tag whereas the first commit was on the RC1 tag.

PR #128, opened 2017-01-18
Restructure conceptual documentation on Beam site.

hadarhg:
Restructure and rename "Documentation" tab on Beam site.


hadarhg:
R: @davorbonaci

davorbonaci:
Random thoughts:
* Visually, I think we have too many top-level menus. Another solution may be nesting (sub-menus inside a menu). It would probably be less of an issue for a vertical menu, as opposed to a horizontal one.
* Some top-level menus have a top-page, while others don't. We should be consistent.
* `Using the documentation` is now under `Concepts`, which makes less sense than before when the title of the menu was `Documentation`.
* Capability matrix should probably be in its own section, as well as the direct runner. Some intro page would definitely be good here.
* Some menu names are verbs while other are nouns. We should be consistent.

I think all this is ~fine, but it is getting less attractive visually. The only blocking issue is consistency.

davorbonaci:
Close?

PR #129, opened 2017-01-21
[BEAM-751] Extract Python code snippets from tested examples.

robertwb:


robertwb:
R: @davorbonaci

davorbonaci:
Thanks @robertwb. This is great.

CC: @hadarhg @melap -- this will make it so much easier for snippets.

PR #130, opened 2017-01-26
Python docs

hadarhg:
- Add Quickstart for Python
- Add Python to WordCount doc

### DIFF
--- a/src/get-started/quickstart-py.md
+++ b/src/get-started/quickstart-py.md
@@ -6,6 +6,8 @@ permalink: /get-started/quickstart-py/
 
 # Apache Beam Python SDK Quickstart
 
+This guide shows you how to set up your Python development environment, get the Apache Beam SDK for Python, and run an example pipeline.
+
 * TOC
 {:toc}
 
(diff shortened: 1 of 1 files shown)


hadarhg:
R: @aaltay

src/get-started/wordcount-example.md
```diff
@@ -60,12 +68,20 @@ You can specify a runner for executing your pipeline, such as the `DataflowRunne
     //   options.setRunner(FlinkRunner.class);
 ```
 
+```py
+{% github_sample /apache/beam/blob/python-sdk/sdks/python/apache_beam/examples/snippets/snippets.py tag:examples_wordcount_minimal_options
+%}```
```
hadarhg: Snippet needs to be updated, see https://issues.apache.org/jira/browse/BEAM-1322

src/get-started/quickstart-py.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Create and activate a virtual environment
+
+A virtual environment is a directory tree containing its own Python distribution. To create a virtual environment, create a directory and run:
+
+```
+virtualenv /path/to/directory
+```
+
+A virtual environment needs to be activated for each shell that is to use it.
```
aaltay: Could you link to `https://virtualenv.pypa.io/en/stable/userguide/#activate-script` for the activated. So users of non-Bash shells know what do (i.e. Windows).

We might even explicitly say for other shells see virtualenv documentation and link.

src/get-started/wordcount-example.md
```diff
@@ -79,52 +95,73 @@ The Minimal WordCount pipeline contains five transforms:
...
+                if (!word.isEmpty()) {
+                    c.output(word);
+                }
+            }
+        }
+    }))
+    ```
+
+    ```py
+    # The Flatmap transform is a simplified version of ParDo.
```
aaltay: Why this comment about the FlatMap is not part of the snippet?

hadarhg: I would have preferred it to be text but we don't have the ability to put text inside the language toggles yet so I put it in as a comment temporarily.

hadarhg:
Thanks, Ahmet. Updated, PTAL. Details for Travis failure [here](https://travis-ci.org/apache/beam-site/builds/195407964). Not sure what to do about that, but it seems like it can be ignored. Wdyt?

aaltay:
LGTM. Thank you!

src/get-started/quickstart-py.md
```diff
@@ -0,0 +1,91 @@
...
+
+[Install pip](https://pip.pypa.io/en/stable/installing/), Python's package manager. Check if you already have `pip` installed by running `pip --version`. After installation, verify that you have `pip` version 7.0.0 or newer. To update `pip`, run the following command:
+
+```
+pip install -U pip
+```
+
+### Install virtualenv
+
+It is recommended that you install a [Python virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/)
```
davorbonaci: I'd simplify -- we don't need to give so many commands. Applies to all steps.

If you don't have XYZ version ABC or newer, install it:
   command to install

davorbonaci:
Merging based on @aaltay's LGTM and a quick skim over it.

PR #121, opened 2017-01-05
Progguide io

hadarhg:
Update I/O section according to reviewer comments in PR https://github.com/apache/beam-site/pull/96

PR #133, opened 2017-01-27
Beam version >0.4.0 does no longer use .named()

tkaymak:


james-woods:
closed, as it is included in #121

davorbonaci:
Yes, but thanks anyways @james-woods!

PR #135, opened 2017-01-30
Use PMC instead of PPMC

jbonofre:
Minor change to reflect the graduation: we don't have PPMC anymore, but PMC.


jbonofre:
R: @davorbonaci

davorbonaci:
LGTM. Merging.

PR #140, opened 2017-02-01
Add page about python type safety and type hints

melap:
R: @aaltay 
@dhalperi - I don't know how to force this page to be adapted for Python so the code blocks always show up. If you view it after toggling for Java, no code blocks show at all. Putting the Adapt for toggle on a Python only page seems clunky. I filed JIRA BEAM-1367 for this. Any ideas?

### DIFF
--- a/src/documentation/sdks/python-type-safety.md
+++ b/src/documentation/sdks/python-type-safety.md
@@ -9,22 +9,22 @@ Python is a dynamically-typed language with no static type checking. Because of 
 
 The Apache Beam SDK for Python uses **type hints** during pipeline construction and runtime to try to emulate the correctness guarantees achieved by true static typing. Additionally, using type hints lays some groundwork that allows the backend service to perform efficient type deduction and registration of `Coder` objects.
 
-Python version 3.5 introduces a module called **typing** to provide hints for type validators in the language. The Beam SDK for Python, based on Python version 2.7, implements a subset of ([PEP 484](https://www.python.org/dev/peps/pep-0484/)) and aims to follow it as closely as possible in its own typehints module.
+Python version 3.5 introduces a module called **typing** to provide hints for type validators in the language. The Beam SDK for Python, based on Python version 2.7, implements a subset of [PEP 484](https://www.python.org/dev/peps/pep-0484/) and aims to follow it as closely as possible in its own typehints module.
 
 ## Benefits of Type Hints
 
 The Beam SDK for Python includes some automatic type checking: for example, some `PTransform`s, such as `Create` and simple `ParDo` transforms, attempt to deduce their output type given their input. However, the Beam cannot infer types in all cases. Therefore, the recommendation is that you declare type hints to aid you in performing your own type checks if necessary.
 
-When you use type hints, the runner raises exceptions during pipeline construction time, rather than runtime. For example, the runner generates an exception if it detects that your pipeline applies mismatched `PTransforms` (where the expected outputs of one transform do not match the expected inputs of the following transform). These exceptions are raised whether your pipeline is executing locally or remotely. Introducing type hints for the `PTransform`s you define allows you to catch potential bugs up front in the local runner, rather than after minutes of execution into a deep, complex pipeline.
+When you use type hints, the runner raises exceptions during pipeline construction time, rather than runtime. For example, the runner generates an exception if it detects that your pipeline applies mismatched `PTransforms` (where the expected outputs of one transform do not match the expected inputs of the following transform). These exceptions are raised at pipeline construction time, regardless of where your pipeline will execute. Introducing type hints for the `PTransform`s you define allows you to catch potential bugs up front in the local runner, rather than after minutes of execution into a deep, complex pipeline.
 
 Consider the following example, in which `numbers` is a `PCollection` of `str` values:
 
-```py
+```
 {% github_sample /apache/beam/blob/master/sdks/python/apache_beam/examples/snippets/snippets_test.py tag:type_hints_missing_define_numbers %}```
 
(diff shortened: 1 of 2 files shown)


src/documentation/sdks/python.md
```diff
@@ -3,8 +3,17 @@ layout: default
 title: "Beam Python SDK"
 permalink: /documentation/sdks/python/
 ---
-# Apache Beam Python SDK _[Under Development]_
+# Apache Beam Python SDK
 
-The Beam Python SDK is currently under development on a feature branch. Would you like to contribute? See the Beam [Work in Progress]({{ site.baseurl }}/contribute/work-in-progress/#feature-branches) page to find out how you can help!
+The Python SDK for Apache Beam provides a simple, powerful API for building both batch and streaming parallel data processing pipelines in Python.
```
aaltay: Python SDK does not support streaming currently.

src/documentation/sdks/python.md
```diff
@@ -3,8 +3,17 @@ layout: default
...
+
+## Get Started with the Python SDK
+
+Get started with the [Beam Programming Guide]({{ site.baseurl }}/learn/programming-guide) to learn the basic concepts that apply to all SDKs in Beam.
+
+Then, follow the [Beam Python SDK Quickstart]({{ site.baseurl }}/get-started/quickstart-py) to set up your Python development environment, get the Beam SDK for Python, and run an example pipeline.
+
+## Python Type Safety
+
+Python is a dynamically-typed language with no static type checking. The Beam SDK for Python uses type hints during pipeline construction and runtime to try to emulate the correctness guarantees achieved by true static typing. [Ensuring Python Type Safety]({{ site.baseurl }}/documentation/sdks/python-type-safety) walks through how to use type hints, which help you to catch potential bugs up front in the local runner, rather than after minutes of execution into a deep, complex pipeline.
```
aaltay: Do we need the final part of this paragraph: 'rather than after minutes of execution into a deep, complex pipeline.' ? I am not opposed it, I am want to understand how much value it adds.

src/documentation/sdks/python-type-safety.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Benefits of Type Hints
+
+The Beam SDK for Python includes some automatic type checking: for example, some `PTransform`s, such as `Create` and simple `ParDo` transforms, attempt to deduce their output type given their input. However, the Beam cannot infer types in all cases. Therefore, the recommendation is that you declare type hints to aid you in performing your own type checks if necessary.
+
+When you use type hints, the runner raises exceptions during pipeline construction time, rather than runtime. For example, the runner generates an exception if it detects that your pipeline applies mismatched `PTransforms` (where the expected outputs of one transform do not match the expected inputs of the following transform). These exceptions are raised whether your pipeline is executing locally or remotely. Introducing type hints for the `PTransform`s you define allows you to catch potential bugs up front in the local runner, rather than after minutes of execution into a deep, complex pipeline.
```
aaltay: Instead of `These exceptions are raised whether your pipeline is executing locally or remotely.`

maybe use `These exceptions are raised at pipeline construction time regardless of where your pipeline will execute.`

src/documentation/sdks/python-type-safety.md
```diff
@@ -0,0 +1,138 @@
...
+However, if you enable runtime type checking, the code passes at pipeline construction and fails at runtime. To enable runtime type checking, set the pipeline option `runtime_type_check` to `True`.
+
+```py
+{% github_sample /apache/beam/blob/master/sdks/python/apache_beam/examples/snippets/snippets_test.py tag:type_hints_runtime_on %}```
+
+## Use of Type Hints in Coders
+
+When your pipeline reads, writes, or otherwise materializes its data, the elements in your `PCollection` need to be encoded and decoded to and from byte strings. Byte strings are used for intermediate storage, for comparing keys in `GroupByKey` operations, and for reading from sources and writing to sinks.
+
+The Beam SDK for Python uses Python's native support for serializing objects, a process called **Pickling**, to serialize user functions. However, using the `PickleCoder` comes with several drawbacks: it is less efficient in time and space, and the encoding used is not deterministic, which hinders distributed partitioning, grouping, and state lookup.
```
aaltay: Pickling does not need to be capitalized.

src/documentation/sdks/python.md
melap: done, removed streaming mention.

src/documentation/sdks/python.md
melap: done, tried to simplify this.

src/documentation/sdks/python-type-safety.md
melap: done

src/documentation/sdks/python-type-safety.md
melap: done

melap:
Incorporated review feedback. Since this is a Python specific page, I also removed py from the code blocks temporarily, so they always show up regardless of how someone has set the Java/Python toggle. Even without syntax highlighting, it seems like a better temporary option (until BEAM-1367 is fixed) than users looking at the page and seeing no code blocks at all. Thoughts?

aaltay:
Python content LGTM. Could we remove toggles from python specific pages?

davorbonaci:
SGTM to proceed without syntax highlighting to avoid toggle issues. Another alternative would be to have an empty Java block saying "This applies to Python SDK only". Ok to proceed either way; just make sure this is fully documented in BEAM-1367 (current state, what needs to happen, etc.).

melap:
OK, proceeding without syntax highlighting, I have updated the JIRA with a tracking note that this page will need to be updated after we have a way to force specific language.

aaltay:
Thank you. LGTM.

PR #141, opened 2017-02-01
Blog: "Media recap of the Apache Beam graduation"

davorbonaci:
R: @francesperry @jbonofre @dhalperi 

@everyone -- please let me know if there are other blogs, articles that I might have missed. Would love to include those as well!


francesperry:
LGTM -- that hits the ones I know about.

PR #143, opened 2017-02-02
[BEAM-1376] Delete jobs due to move to main repo.

jasonkuster:
Signed-off-by: Jason Kuster <jasonkuster@google.com>


jasonkuster:
We need to leave the python script in at least -- realized it needs to live in this repo since this is still what is checked out, even though the jobs live in the other repo. All jobs need to live in same repo though, so that is all that can stay.

jasonkuster:
Added it back -- we should be good to merge once https://github.com/apache/beam/pull/1897 is in.

PR #149, opened 2017-02-08
[BEAM-25,BEAM-27] Update the capability matrix for state and timers

kennknowles:
The clarity of the model spec for state and timers are at parity with other model features, and both state and timers are partially supported by the Flink and Dataflow runners.


kennknowles:
(this describes HEAD, not 0.5.0)

PR #154, opened 2017-02-15
Clarify state further in capability matrix

kennknowles:
Some feedback after the blog post was that the term "Keyed State" in the capability matrix is lacking a referent. Basically I got more queries like [this one](https://lists.apache.org/thread.html/7ab688401796e1d7411c9975581ea5001a8c67fa415fff9f720753c9@%3Cdev.beam.apache.org%3E) but actually in response to a whole blog post describing it!

I had thought this term came from Millwheel so there might be some use in tying the two together. But it is actually [not in the paper](https://www.cs.cmu.edu/~pavlo/courses/fall2013/static/papers/p734-akidau.pdf). I think it is used a bit in Flink. A more general terminology seems appropriate for the capability matrix here.

In this PR I have suggested a general terminology, but there may be better options.

### DIFF
--- a/src/_data/capability-matrix.yml
+++ b/src/_data/capability-matrix.yml
@@ -196,7 +196,7 @@ categories:
             l2: Not implemented in runner.
             l3: 
 
-      - name: Keyed State
+      - name: Stateful Processing
         values:
           - class: model
             l1: 'Yes'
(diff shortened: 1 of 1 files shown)


davorbonaci:
LGTM. Merging.

PR #155, opened 2017-02-15
Fix typo in stateful processing post

kennknowles:


kennknowles:
Since it is just adding a period, I'll go ahead and merge this myself right away.

PR #156, opened 2017-02-15
Fix some typos and small formatting issuesin the last stateful processing blog post

iemejia:
- The first image in the section "Example: arbitrary-but-consistent index assignment" is really hard to see with the default settings. I made it bigger but probably it would be better to change the proportions of the elements of the diagram.
- There is a small typo I think (uninterseting => uninteresting).
- "You can provide the opportunity for parallelism by making sure that table has enough columns, either via many keys in few windows - for example, a globally windowed stateful computation keyed by user ID - or via many windows over few keys - for example, a fixed windowed stateful computation over a global key. Caveat: all Beam runners today parallelize only over the key." I suppose there is an issue with the list here in the markdown, or at least it seems more clear like this I think.

### DIFF
--- a/src/_posts/2017-02-13-stateful-processing.md
+++ b/src/_posts/2017-02-13-stateful-processing.md
@@ -196,7 +196,7 @@ want to write a transform that maps input to output like this:
 <img class="center-block" 
     src="{{ site.baseurl }}/images/blog/stateful-processing/assign-indices.png"
     alt="Assigning arbitrary but unique indices to each element"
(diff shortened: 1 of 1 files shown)


iemejia:
R: @kennknowles

kennknowles:
I agree with your edits, and thanks for this! I'll touch up some punctuation and merge it.

iemejia:
Great, thanks to you for the post, it is really good.

PR #159, opened 2017-02-17
Add Apache in front of initial Beam reference

melap:
R: @davorbonaci @aaltay 
Changes to address PR 157 comments. Removed Beam from title to be consistent with other Python pages, and add Apache in front of initial Beam reference.

### DIFF
--- a/src/documentation/sdks/python-custom-io.md
+++ b/src/documentation/sdks/python-custom-io.md
@@ -43,7 +43,7 @@ It is critical to exhaustively unit-test all of your `Source` and `Sink` subclas
 You can use test harnesses and utility methods available in the [source_test_utils module](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/source_test_utils.py) to develop tests for your source.
 
 
-## <a name="creating-sources"></a>Creating a Custom Source
+## <a name="creating-sources"></a>Creating a New Source
 
 You should create a new source if you'd like to use the advanced features that the Source API provides:
 
@@ -157,7 +157,7 @@ To create a source for a new file type, you need to create a sub-class of `FileB
 See [AvroSource](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/avroio.py) for an example implementation of `FileBasedSource`.
 
 
-## <a name="reading-sources"></a>Reading from a Custom Source
+## <a name="reading-sources"></a>Reading from a New Source
 
 The following example, `CountingSource`, demonstrates an implementation of `BoundedSource` and uses the SDK-provided `RangeTracker` called `OffsetRangeTracker`.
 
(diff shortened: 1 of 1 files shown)


melap:
I'm not sure about "Python SDK", we're need to say "Custom Sinks and Sources in Apache Beam SDK for Python" which seems long. Thoughts?

src/documentation/sdks/python-custom-io.md
```diff
@@ -1,11 +1,11 @@
 ---
 layout: default
-title: "Beam Custom Sources and Sinks for Python"
+title: "Custom Sources and Sinks for Python"
```
davorbonaci: > I'm not sure about "Python SDK", would we need to say "Custom Sources and Sinks in Apache Beam SDK for Python" which seems long? Thoughts? Or could axe Python from the title entirely. Or "Python Custom Sources and Sinks"?

Ack on this concern; we shouldn't make it too long.

I'd argue for:
* The page title (the thing that goes into `title:`) should (almost) always be `Apache Beam XXX`. Very rarely `XXX` is fine, but it shouldn't be `Beam XXX`.
* The H1 heading (the thing that goes into `#`) can be `XXX` or `Apache Beam XXX`, never `Beam XXX`. Repeating Beam is not necessary (given that context is already established with the logo, menus, etc.).
* Avoid mixing "Python" and "Python SDK". These aren't sources for Python; they are sources in the Beam's Python SDK -- they don't apply generally to Python. We colloquially use "Python" all the time, but in writing should almost always be "Python SDK".

melap: Got it. Made some changes, PTAL.

melap:
Updated from feedback and changed "custom" to "new" throughout the page.

src/documentation/sdks/python-custom-io.md
```diff
@@ -45,7 +45,7 @@ You can use test harnesses and utility methods available in the [source_test_uti
 
 ## <a name="creating-sources"></a>Creating a Custom Source
```
davorbonaci: There are several more `custom` words on this page -- perhaps clean up all of them?

melap:
My bad, I did a search but must have been case sensitive. Updated 4 additional section titles. I left the page name (it contains "custom") alone for now to not cause more churn, as once we have a unified "Creating new IO" page that contains both Python and Java tabs, this page will become a redirect anyway. If you'd prefer it updated anyway, let me know.

src/documentation/sdks/python-custom-io.md
melap: Fixed 4 additional "custom" references in titles.

melap:
Good point. When I create the unified version, I'll clean that up.

PR #160, opened 2017-02-22
Fix broken links due to code path changes

melap:


davorbonaci:
Also, @aaltay @sb2nov for fyi.

aaltay:
FYI: We are planning to rename `io/google_cloud_platform/` to `io/gcp`. We will update this if that happens.

PR #161, opened 2017-02-22
Fix broken links due to directory renames in Beam

sb2nov:
Ref: https://github.com/apache/beam/pull/2079

R: @aaltay PTAL


aaltay:
LGTM. I can merge this after https://github.com/apache/beam/pull/2079

PR #162, opened 2017-02-23
[BEAM-1499] Adding testing related info: TestPipeline, API surface.

staslev:
R: @jasonkuster

### DIFF
--- a/src/contribute/testing.md
+++ b/src/contribute/testing.md
@@ -371,3 +371,158 @@ RunnableOnService tests are tests built to use the Beam TestPipeline class,
 which enables test authors to write simple functionality verification. They are
 meant to use some of the built-in utilities of the SDK, namely PAssert, to
 verify that the simple pipelines they run end in the correct state.
+
+
+### Effective use of the TestPipeline JUnit rule
+
+`TestPipeline` is JUnit rule designed to facilitate testing pipelines. 
+In combination with `PAssert`, the two can be used for testing and 
+writing assertions over pipelines. However, in order for these assertions 
+to be effective, the constructed pipeline **must** be run by a pipeline 
+runner. If the pipeline is not run (i.e., executed) then the 
+constructed `PAssert` statements will not be triggered, and will thus 
+be ineffective. 
+
+To prevent such cases, `TestPipeline` has some protection mechanisms in place.
+
+__Abandoned node detection (performed automatically)__
+
+Abandoned nodes are `PTransforms`, `PAsserts` included, that were not 
+executed by the pipeline runner. Abandoned nodes are most likely to occur 
+due to the one of the following scenarios:
+ 1. Lack of a `pipeline.run()` statement at the end of a test. 
+ 2. Addition of `PTransform`s  after the pipeline has already run.
+
+Abandoned node detection is *automatically enabled* when a real pipeline 
+runner (i.e. not a `CrashingRunner`) and/or a 
+`@NeedsRunner` / `@RunnableOnService` annotation are detected. 
+
+Consider the following test:
+  
+```java
+// Note the @Rule annotation here
+@Rule
+public final transient TestPipeline pipeline = TestPipeline.create();
+
+@Test
+@Category(NeedsRunner.class)
+public void myPipelineTest() throws Exception {
+
+final PCollection<String> pCollection = 
+  pipeline
+    .apply("Create", Create.of(WORDS).withCoder(StringUtf8Coder.of()))
+    .apply(
+        "Map1",
+        MapElements.via(
+            new SimpleFunction<String, String>() {
+
+              @Override
+              public String apply(final String input) {
+                return WHATEVER;
+              }
+            }));
+            
+PAssert.that(pCollection).containsInAnyOrder(WHATEVER);       
+
+/* ERROR: pipeline.run() is missing, PAsserts are ineffective */
+}
+```
+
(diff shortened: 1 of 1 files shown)


staslev:
R: @jasonkuster

src/contribute/testing.md
```diff
@@ -371,3 +371,135 @@ RunnableOnService tests are tests built to use the Beam TestPipeline class,
...
 meant to use some of the built-in utilities of the SDK, namely PAssert, to
 verify that the simple pipelines they run end in the correct state.
+
+
+### Effective use of `TestPipeline`
+
+`TestPipeline` and `PAssert` are great tools for testing and writing assertions
+ over pipelines. However, in order for these tools to be effective, the 
+ constructed pipeline **must** be run by a pipeline runner. 
+If the pipeline is no run (i.e., executed) the the constructed `PAssert`
```
ssisk: no -> not
the -> then

src/contribute/testing.md
```diff
@@ -371,3 +371,135 @@ RunnableOnService tests are tests built to use the Beam TestPipeline class,
...
+classes, Beam provides the `ApiSurface` utility class. 
+Using the `ApiSurface` class,  we can assert the API surface against an 
+expected set of classes.
+
+Consider the following snippet:
+```java
+  @Test
+  public void testMyApiSurface() throws Exception {
+
+	final Package thisPackage = getClass().getPackage();
```
ssisk: I'm not sure if the formatting matters in this context, but if so: this line is different from the next one.

src/contribute/testing.md
```diff
@@ -371,3 +371,135 @@ RunnableOnService tests are tests built to use the Beam TestPipeline class,
...
+  The application of the `read` transform throws an exception, which is then 
+  handled by the `thrown` `ExpectedException` rule. 
+  In light of this exception, the fact this test has abandoned nodes 
+  (the `read` transform) does not play a role since the test fails before 
+  the pipeline would have been executed (had there been a `run()` statement). 
+   
+ * *Auto-add `run()`* (disabled by default)
+   A `TestPipeline` instance can be configured to auto-add a missing `run()` 
+   statement by setting `testPipeline.enableAutoRunIfMissing(true/false)`. 
+   If this feature is enabled, the snippet above will not throw an exception,
```
ssisk: "snippet above" is a bit misleading since the next snippet up from this comment is about the ExpectedException stuff.

src/contribute/testing.md
```diff
@@ -371,3 +371,135 @@ RunnableOnService tests are tests built to use the Beam TestPipeline class,
...
+  handled by the `thrown` `ExpectedException` rule. 
+  In light of this exception, the fact this test has abandoned nodes 
+  (the `read` transform) does not play a role since the test fails before 
+  the pipeline would have been executed (had there been a `run()` statement). 
+   
+ * *Auto-add `run()`* (disabled by default)
+   A `TestPipeline` instance can be configured to auto-add a missing `run()` 
+   statement by setting `testPipeline.enableAutoRunIfMissing(true/false)`. 
+   If this feature is enabled, the snippet above will not throw an exception, 
+   and the pipeline will get executed.
```
ssisk: The background info on why this is necessary is very helpful. Thanks!

Further thoughts/hopes: 
1) If the idea is that this is the only explanation of how to use Test Pipeline, I think it needs more fleshing out of "how do I actually use this?" For example, you show @Rule usage, but don't actually mention that as something you need to do (ie, it would be easy to miss that it's in there/important.) What are the expectations on a user who wants to use TestPipeline now?
2) As I mentioned in BEAM-1543, the javadocs are out of date - are you planning on updating those at some point? What belongs here vs in the javadocs for TestPipeline?

As an example of #1: 
I don't understand is what the @Rule is doing. I just went and read the junit docs on @Rule (I had to figure out which @Rule that is - it's not obvious from this text that it's the junit rule) and I now kind of get why Rule is what we're using here, but I think a bit more explanation would be helpful.

src/contribute/testing.md
staslev: 👍

src/contribute/testing.md
staslev: 👍

src/contribute/testing.md
staslev: Thanks for the feedback, I'll make sure to update this section and port the relevant info to the javadoc.

ssisk:
Your changes look like they addressed my suggestions. Thanks!

src/contribute/testing.md
```diff
@@ -371,3 +371,146 @@ RunnableOnService tests are tests built to use the Beam TestPipeline class,
...
+`pipeline` is never run, making this test ineffective. If this test method 
+is run using an actual pipeline runner, an exception will be thrown 
+indicating that there was no `run()` invocation in the test.
+
+Exceptions that are thrown prior to executing a pipeline, will fail 
+the test unless handled by an `ExpectedException` rule.
+
+Consider the following test:  
+  
+```java
```
davorbonaci: Due to Javascript issues (unrelated to this PR), it is possible that the browser doesn't render the code snippet if Python snippet was last selected on another page.

To work around the problem, I suggest to add an empty Python snippet below (see example in Kenn's state blog post).

staslev: Thanks for the heads up, on it!

PR #163, opened 2017-02-24
Fix the quickstart documentation

sb2nov:
R: @aaltay PTAL


aaltay:
LGTM. We should eventually have tabs here for DirectRunner and DataflowRunner separately.

aaltay:
https://issues.apache.org/jira/browse/BEAM-1549 for updating the quickstart with tabs.

PR #166, opened 2017-02-24
[BEAM-1552] Add contrib guide comments about using squash & force push.

ssisk:
I encountered some issues while starting to work on Beam and thought I'd update the contributor's guide.

The contrib guide contains pretty detailed git commands, but:
1. For pushing to your branch, the git commands don't actually work when you try to use them in real life (you need to force push)
2. For squashing, we tell contributors that they *can* squash, but we don't actually say that we expect them to.

### DIFF
--- a/src/contribute/contribution-guide.md
+++ b/src/contribute/contribution-guide.md
@@ -186,9 +186,9 @@ Periodically while you work, and certainly before submitting a pull request, you
 
 Remember to always use `--rebase` parameter to avoid extraneous merge commits.
 
-To push your local, committed changes to your (forked) repository on GitHub, run:
+Then you can push your local, committed changes to your (forked) repository on GitHub. Since rebase may change that branch's history, you may need to force push. You'll run:
 
-	$ git push <GitHub_user> <my-branch>
+	$ git push <GitHub_user> <my-branch> --force
 
 ### Testing
 All code should have appropriate unit testing coverage. New code should have new tests in the same contribution. Bug fixes should include a regression test to prevent the issue from reoccurring.
@@ -201,9 +201,11 @@ For contributions to the Java code, run unit tests locally via Maven. Alternativ
 Once the initial code is complete and the tests pass, it’s time to start the code review process. We review and discuss all code, no matter who authors it. It’s a great way to build community, since you can learn from other developers, and they become familiar with your contribution. It also builds a strong project by encouraging a high quality bar and keeping code consistent throughout the project.
 
 ### Create a pull request
-Organize your commits to make your reviewer’s job easier. Use the following command to re-order, squash, edit, or change description of individual commits.
+Organize your commits to make your reviewer’s job easier. Reviewers normally prefer multiple small pull requests, instead of a single large pull request. Within a pull request, a relatively small number of commits that break the problem into logical steps is preferred. For most pull requests, you'll squash your changes down to 1 commit. You can use the following command to re-order, squash, edit, or change description of individual commits.
 
     $ git rebase -i origin/master
+
+You'll then push to your branch on GitHub. Note: when updating your commit after pull request feedback and use squash to get back to one commit, you will need to do a force submit to the branch on your repo.
 
 Navigate to the [Beam GitHub mirror](https://github.com/apache/beam) to create a pull request. The title of the pull request should be strictly in the following format:
 


ssisk:
@davorbonaci

src/contribute/contribution-guide.md
```diff
@@ -201,10 +201,12 @@ For contributions to the Java code, run unit tests locally via Maven. Alternativ
 Once the initial code is complete and the tests pass, it’s time to start the code review process. We review and discuss all code, no matter who authors it. It’s a great way to build community, since you can learn from other developers, and they become familiar with your contribution. It also builds a strong project by encouraging a high quality bar and keeping code consistent throughout the project.
 
 ### Create a pull request
-Organize your commits to make your reviewer’s job easier. Use the following command to re-order, squash, edit, or change description of individual commits.
+Organize your commits to make your reviewer’s job easier. Reviewers normally prefer a minimum number of commits that break up a PR into logical steps. For most PRs, you'll squash your changes down to 1 commit. You can use the following command to re-order, squash, edit, or change description of individual commits.
 
     $ git rebase -i origin/master
 
+You'll then push to your branch on github. Note: when updating your commit after PR feedback and use squash to get back to one commit, you will need to do a force submit to the branch on your repo.
```
davorbonaci: github -> GitHub

But, I'd remove this paragraph altogether. It is ok not to squash here. Or, just squash at the very end.

src/contribute/contribution-guide.md
```diff
@@ -186,9 +186,9 @@ Periodically while you work, and certainly before submitting a pull request, you
 
 Remember to always use `--rebase` parameter to avoid extraneous merge commits.
 
-To push your local, committed changes to your (forked) repository on GitHub, run:
+Then you can push your local, committed changes to your (forked) repository on GitHub. Since rebase changes that branch's history, you'll need to force push. You'll run:
```
davorbonaci: changes -> may change

davorbonaci: you'll need -> you may need

src/contribute/contribution-guide.md
```diff
@@ -201,10 +201,12 @@ For contributions to the Java code, run unit tests locally via Maven. Alternativ
 Once the initial code is complete and the tests pass, it’s time to start the code review process. We review and discuss all code, no matter who authors it. It’s a great way to build community, since you can learn from other developers, and they become familiar with your contribution. It also builds a strong project by encouraging a high quality bar and keeping code consistent throughout the project.
 
 ### Create a pull request
-Organize your commits to make your reviewer’s job easier. Use the following command to re-order, squash, edit, or change description of individual commits.
+Organize your commits to make your reviewer’s job easier. Reviewers normally prefer a minimum number of commits that break up a PR into logical steps. For most PRs, you'll squash your changes down to 1 commit. You can use the following command to re-order, squash, edit, or change description of individual commits.
```
davorbonaci: Points:
* multiple small PRs, instead of a large PR
* within a PR, a relatively small number of commits that break the problem into logical steps

Everwhere "PR" -> pull request.

src/contribute/contribution-guide.md
ssisk: updated github text.

for other text: 
This information about needing to force push is non-trivial to discover for new git users. General guidance given in git documentation is "force pushing is unusual/bad".  

Generaly, if we think it's necessary to provide the git commands to do things, then I think this information is consistent with providing that level of detail. 

If we didn't provide any of the exact commands (implying that we assume our users already know how to use git), then I would agree with you that this is unnecessary info.

src/contribute/contribution-guide.md
ssisk: done

ssisk: done

src/contribute/contribution-guide.md
ssisk: done. I think I incorporated this correctly.

src/contribute/contribution-guide.md
ssisk: Let's say I have a PR with 2 logical steps to it, changing a medium sized set of files. 

I break my initial PR up into 2 commits, each addressing a subset of the files. 

I then need to react to PR feedback. If I need to change the files in both commits, is there a way to correctly squash the changes so that the files I change are separated out into the correctly commits? 

If they don't squash the changes, then we get a bunch of commits like "updating based on PR feedback", so obviously that's not right either.

If there isn't a good way to do this, then inevitably 1 of the 2 commits ends up getting a bunch of the random fixes, right? So why shouldn't we just be telling people to do 1 commit?

ssisk:
The build failure isn't mine - I filed a JIRA for it.

PR #167, opened 2017-02-27
[BEAM-1559] Reference HBase support in the documentation

tom-haines:
References to the HBase IO connector recently added are updated into the documentation.


jbonofre:
R: @jbonofre

PR #168, opened 2017-03-02
Add windowing section to programming guide

melap:
R: @kennknowles


src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
 {:.language-py}
 The Beam SDK for Python does not support annotating data types with a default coder. If you would like to set a default coder, use the method described in the previous section, *Setting the default coder for a type*.
 
-<a name="windowing"></a>
-<a name="triggers"></a>
+## <a name="windowing"></a>Working with windowing
+
+Windowing subdivides a `PCollection` according to the timestamps of its individual elements. Beam transforms that aggregate multiple elements, such as `GroupByKey` and `Combine`, work implicitly on a per-window basis—that is, they process each `PCollection` as a succession of multiple, finite windows, though the entire collection itself may be of unbounded size.
```
kennknowles: nit: I initially read "Beam transforms" as subject-verb instead of as attributive noun - plural, then my brain sorted it out a few words later

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+
+A related concept, called **triggers**, determines when to emit the results of aggregation as unbounded data arrives. Using a trigger can help to refine the windowing strategy for your `PCollection` to deal with late-arriving data or to provide early results. See the [triggers](#triggers) section for more information.
+
+### Windowing basics
+
+Windowing is most useful with an unbounded `PCollection`, which represents a continuously updating data set of unknown, and possibly infinite size (e.g. streaming data). Some Beam transforms, such as `GroupByKey` and `Combine`, group multiple elements by a common key. Ordinarily, that grouping operation groups all of the elements that have the same key within the entire data set. With an unbounded data set, it is impossible to collect all of the elements, since new elements are constantly being added and may be infinitely many.
```
kennknowles: Might soften to `especially useful with an unbounded PCollection`. But now that I read this and the following paragraph, I wonder if you might lead with `Some Beam transforms... group multiple elements by a common key` and then the current lead sentence (in spirit, not exact text) comes after `... inifinitely many.`

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+
+In the Beam model, any `PCollection` can be subdivided into logical windows. Each element in a `PCollection` is assigned to one or more windows according to the `PCollection`'s windowing function, and each individual window contains a finite number of elements. Grouping transforms then consider each `PCollection`'s elements on a per-window basis. `GroupByKey`, for example, implicitly groups the elements of a `PCollection` by _key and window_.
+
+**Caution:** Default windowing behavior is to assign all elements of a `PCollection` to a single, global window, _even for unbounded `PCollection`s_. Before you use a grouping transform such as `GroupByKey` on an unbounded `PCollection`, you must set a non-global windowing function. See [Setting your PCollection's windowing function](#setwindowingfunction).
```
kennknowles: Not pedantically true, because you can also set a non-default trigger while leaving the global window in place. I see the following paragraph makes that addendum. I wonder if there is a way to flow better in case someone stops reading at the first paragraph?

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+*  Sliding Time Windows
+*  Per-Session Windows
+*  Single Global Window
+*  Calendar-based Windows (not supported by the Beam SDK for Python)
+
+Note that each element can logically belong to more than one window, depending on the windowing function you use. Sliding time windowing, for example, creates overlapping windows wherein a single element can be assigned to multiple windows.
+
+#### Fixed time windows
+
+The simplest form of windowing is using **fixed time windows**: given a timestamped `PCollection` which might be continuously updating, each window might capture (for example) five minutes worth of elements.
```
kennknowles: Somewhere around here, in case the user has paged out the processing time vs. event time distinction, it might be worth recalling it. Particularly `five minutes worth of elements` sounds a bit like processing time, versus something like `might capture all elements with timestamps that fall into a five minute interval`.

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+
+A fixed time window represents a consistent duration, non overlapping time interval in the data stream. Consider windows with a five-minute duration: all of the elements in your unbounded `PCollection` with timestamp values from 0:00:00 up to (but not including) 0:05:00 belong to the first window, elements with timestamp values from 0:05:00 up to (but not including) 0:10:00 belong to the second window, and so on.
+
+![Diagram of fixed time windows, 30s in duration]({{ "/images/fixed-time-windows.png" | prepend: site.baseurl }} "Fixed time windows, 30s in duration")
+
+**Figure:** Fixed time windows, 30s in duration.
+
+#### Sliding time windows
+
+A **sliding time window** also represents time intervals in the data stream; however, with sliding time windowing, the windows overlap. For example, each window might capture five minutes worth of data, but a new window starts every ten seconds. The frequency with which sliding windows begin is called the _period_. Therefore, our example would have a window _duration_ of five minutes and a _period_ of ten seconds.
```
kennknowles: Nit about `windows` versus `windowing` just in terms of going for extreme uniformity of use: Maybe `however, sliding time windows can overlap`.

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+
+Because multiple windows overlap, most elements in a data set will belong to more than one window. This kind of windowing is useful for taking running averages of data; using sliding time windows, you can compute a running average of the past five minutes' worth of data, updated every ten seconds, in our example.
+
+![Diagram of sliding time windows, with 1 minute window duration and 30s window period]({{ "/images/sliding-time-windows.png" | prepend: site.baseurl }} "Sliding time windows, with 1 minute window duration and 30s window period")
+
+**Figure:** Sliding time windows, with 1 minute window duration and 30s window period.
+
+#### Session windows
+
+A **session window** function defines windows that contain elements that are within a certain gap duration of another element. Session windowing applies on a per-key basis and is useful for data that is irregularly distributed with respect to time. For example, a data stream representing user mouse activity may have long periods of idle time interspersed with high concentrations of clicks. If data arrives after the minimum specified gap duration time, this initiates the start of a new window.
```
kennknowles: In contrast, I think the use of `Session windowing` here is perfect, and `Session windows` would not be right. So maybe I am getting way too deep into word choices that are roughly equivalent.

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+
+You can use a single global window if you are working with an unbounded data set, e.g. from a streaming data source; however, use caution when applying aggregating transforms such as `GroupByKey` and `Combine`. A single global window with a default trigger generally requires the entire data set to be available before processing, which is not possible with continuously updating data. To perform aggregations on an unbounded `PCollection` that uses global windowing, you should specify a non-default trigger for that `PCollection`.
+
+### <a name="setwindowingfunction"></a>Setting your PCollection's windowing function
+
+You can set the windowing function for a `PCollection` by applying the `Window` transform. When you apply the `Window` transform, you must provide a `WindowFn`. The `WindowFn` determines the windowing function your `PCollection` will use for subsequent grouping transforms, such as a fixed or sliding time window.
+
+Beam provides pre-defined `WindownFn`s for the basic windowing functions. You can also define your own `WindowFn` if you have a more complex need.
```
kennknowles: `the basic windowing functions describe here`?

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+        Window.<String>into(new GlobalWindows()));
+```
+```py
+{% github_sample /apache/beam/blob/master/sdks/python/apache_beam/examples/snippets/snippets_test.py tag:setting_global_window
+%}
+```
+
+### Time skew, data lag, and late data
+
+In any data processing system, there is a certain amount of lag between the time a data event occurs (the "event time", determined by the timestamp on the data element itself) and the time the actual data element gets processed at any stage in your pipeline (the "processing time", determined by the clock on the system processing the element). In addition, there are no guarantees that data events will appear in your pipeline in the same order that they were generated in various places on the web.
```
kennknowles: `various places on the web` is just an example, eh?

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+
+For example, let's say we have a `PCollection` that's using fixed-time windowing, with windows that are five minutes long. For each window, Beam must collect all the data with an _event time_ timestamp in the given window range (between 0:00 and 4:59 in the first window, for instance). Data with timestamps outside that range (data from 5:00 or later) belong to a different window.
+
+However, data isn't always guaranteed to arrive in a pipeline in correct time order, or to always arrive at predictable intervals. Beam tracks a _watermark_, which is the system's notion of when all data in a certain window can be expected to have arrived in the pipeline. Data that arrives with a timestamp after the watermark is considered **late data**.
```
kennknowles: I'd scratch `correct` from `time order`. Timestamp order just is what it is, and it isn't what you get in Beam :-)

src/documentation/programming-guide.md
```diff
@@ -1283,7 +1283,225 @@ public class MyCustomDataType {
...
+              .withAllowedLateness(Duration.standardDays(2)));
+```
+
+When you set `.withAllowedLateness` on a `PCollection`, that allowed lateness propagates forward to any subsequent `PCollection` derived from the first `PCollection` you applied allowed lateness to. If you want to change the allowed lateness later in your pipeline, you must do so explictly by applying `Window.withAllowedLateness()` again.
+
+You can also use triggers to help you refine the windowing strategy for a `PCollection`. You can use triggers to determine exactly when each individual window aggregates and reports its results, including how the window emits late elements.
+
+### Adding timestamps to a PCollection's elements
+
+An unbounded source provides a timestamp for each element. Depending on your unbounded source, you may need to configure how the timestamp is extracted from the raw data stream.
```
kennknowles: Just want to comment that this section is extremely useful standalone, even though just a recapitulation. Nice.

src/documentation/programming-guide.md
melap: Modified this a bit, ptal.

src/documentation/programming-guide.md
melap: I did some reformatteing so it's presented as a "you must do at least one of the following, set non-global windowing function or  set a non-default trigger", let me know if this still isn't quite clear.

melap:
Thanks for your great feedback. Made changes, and updated the two diagrams (fixed and sliding windows) to not have broken up boxes. Tried to make the sliding window outlines more visible also. Let me know if there are additional changes you think would be beneficial.

davorbonaci:
@kennknowles, another look perhaps?

davorbonaci:
@kennknowles, PTAL?

kennknowles:
This is superb. LGTM

PR #170, opened 2017-03-04
Fix possible erranous naming conventions reference

aviemzur:


aviemzur:
R: @jkff 

I might be wrong, but I think these references for naming conventions are more clear cut and cause less confusion.

https://en.wikipedia.org/wiki/Camel_case

PR #173, opened 2017-03-09
[BEAM-1665] Add Pipeline I/O section to website - outline + move some existing txt

ssisk:
This PR has a fair number of TODOs in the website content, but I have PRs queued up with more content from the IO authoring guide and none of the TODOs are for content already on the website. 

It has very little actual new content - mostly, it created structure and moves existing content around.

* I did not to go with a single page for all this content b/c both java and python have enough unique content that they deserve their own separate sections (ie, just tabs on the code isn't enough), and the "click to the next page" model currently implemented allows the user to pick java vs python, but then after reading those pages, the next page for both points at the same place - the users mostly follow the same path, but for java vs python specific content, they will diverge then converge again.
* I moved the "list of built-in I/O" content over to it's own separate page since it'd be nice to have more content there - e.g. capabilities matrix, and it felt special enough to pull out of the programming guide.
* We decided not to put all of this content in the contribute section of the site since the expectation is we don't think all users will contribute their IO transforms, so we want most of the docs to just be about writing an IO transforms, and they lay out the expectations in the contribute part of the IO section.

R @melap

### DIFF
--- a/src/_includes/header.html
+++ b/src/_includes/header.html
@@ -42,7 +42,7 @@
               <li><a href="{{ site.baseurl }}/documentation/pipelines/design-your-pipeline/">Design Your Pipeline</a></li>
               <li><a href="{{ site.baseurl }}/documentation/pipelines/create-your-pipeline/">Create Your Pipeline</a></li>
               <li><a href="{{ site.baseurl }}/documentation/pipelines/test-your-pipeline/">Test Your Pipeline</a></li>
-              <li> <a href="{{ site.baseurl }}/documentation/io/io-toc/">Pipeline I/O</a></li>
+              <li><a href="{{ site.baseurl }}/documentation/io/io-toc/">Pipeline I/O</a></li>
               <li role="separator" class="divider"></li>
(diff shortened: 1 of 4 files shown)


melap:
Thanks, LGTM as a skeleton starting point. Looking forward to seeing it full of good stuff.

ssisk:
Thanks for the review Melissa! 
cc @davorbonaci @francesperry for merge (and review if you'd like)

ssisk:
sure, I'll take a pass and see what I can comment out so that it doesn't present a ton of TODOs to the users. 

That'll probably look like: initially the TOC page will be there and the "using" section will be visible, but not the authoring section.

ssisk:
okay! I've:
1. Updated the TOC so that it only displays one "this is a work in progress" note for the authoring content, and users won't see the authoring content yet/the other set of TODOs.
2. Looked around to see if there's any pointers directly into the "authoring transforms" content, and changed the one reference (in the programming guide) to point at the IO TOC instead of directly at the content.

PTAL

PR #184, opened 2017-03-17
Add jkff as committer

jkff:
R: @davorbonaci

### DIFF
--- a/src/_beam_team/team.md
+++ b/src/_beam_team/team.md
@@ -68,6 +68,12 @@ members:
     organization: Google
     roles: committer
     time_zone: "-8"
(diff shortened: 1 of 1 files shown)


src/_beam_team/team.md
```diff
@@ -122,4 +122,10 @@ members:
     organization: Slack
     roles: committer, PMC
     time_zone: "-8"
+  - name: Eugene Kirpichov
```
davorbonaci: Can you please sort it alphabetically by last name?

jkff: Done. I wasn't sure how it was sorted; somehow didn't think of last name :)

PR #189, opened 2017-03-18
Improve committing website changes section in the contribution guide

iemejia:
R: @aviemzur @davorbonaci

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -52,3 +52,6 @@ kramdown:
 release_latest: 0.6.0
 
 # Plugins are configured in the Gemfile.
+
+# Set the time zone for site generation, fixed to US Pacific Time
+timezone: America/Los_Angeles


src/contribute/contribution-guide.md
```diff
@@ -458,8 +458,15 @@ When you are ready, submit a pull request using the [Beam Site GitHub mirror](ht
...
 
-#### Committing website changes
+#### Committing website changes (committers only)
 
 Follow the same committer process as above, but using repository `apache/beam-site` and branch `asf-site`.
 
-In addition, the committer is responsible for doing the final `jekyll build` to generate the static content, so follow the instructions above to install `jekyll`.
+In addition, the committer is responsible for doing the final `bundle exec jekyll build` to generate the static content, so follow the instructions above to install `jekyll`.
+
+This command generates the `content/` directory. The committer should add and commit the content related to the PR. Pay attention to add only the files related to the given pull request since `jekyll` regenerates all the website and sometimes there are unrelated changes e.g. Updated timestamps for old blog posts that should not be added to the website repository among others.
```
davorbonaci: I think updated timestamps are a sign of a bug in the configuration. @aljoscha found out this too. Most of us are in the Pacific timezone, so we don't see this problem. What happens is:
* Jekyll uses the timezone on the local machine to generate timestamps
* If the last time website was generated was performed on a machine in the same time zone, there are no spurious changes.
* If the last time website was generated was performed on a machine in a different time zone, all timestamp appear to have changed.

I think we should just find a way to reconfigure Jekyll to use a given time zone, regardless of who generates the website. And, this problem should just go away. Possibly as simple as `timezone` parameter in the [Jekyll configuration](https://jekyllrb.com/docs/configuration/). That might be a better change than this explanation...

iemejia: You are right Davor, much better solution, I am going to fix the timezone to America/Los_Angeles because this is the default one so it won't clash with the timestamps. I will add this to the PR.

iemejia:
It should be OK now. Please review.

PR #193, opened 2017-03-22
Provide correct runner option for GCD

meunierd:
Passing `dataflow` actually errors out.


src/documentation/runners/dataflow.md
```diff
@@ -72,7 +72,7 @@ When executing your pipeline with the Cloud Dataflow Runner, set these pipeline
 <tr>
   <td><code>runner</code></td>
   <td>The pipeline runner to use. This option allows you to determine the pipeline runner at runtime.</td>
-  <td>Set to <code>dataflow</code> to run on the Cloud Dataflow Service.</td>
```
davorbonaci: Both 'dataflow' and 'dataflowrunner', and regardless of casing, should work. If not, that's a code bug, so I'd leave this as-is.

PR #198, opened 2017-03-31
Update Capability Matrix: Flink supports state/timers in batch mode

aljoscha:
This was implemented a while back for batch mode.

R: @davorbonaci


davorbonaci:
Also, CC: @kennknowles.

aljoscha:
Thanks! 😃

PR #200, opened 2017-04-03
Fix the link to the apache ICLA

sb2nov:
R: @davorbonaci @aaltay PTAL


sb2nov:
The text link is giving a 404 so updated it to the PDF based on https://www.apache.org/licenses/#clas

aaltay:
LGTM

PR #201, opened 2017-04-03
[BEAM-1711] Add information about extra requires in Python Quickstart

sb2nov:
R: @aaltay PTAL

### DIFF
--- a/src/get-started/quickstart-py.md
+++ b/src/get-started/quickstart-py.md
@@ -68,6 +68,24 @@ Install the latest Python SDK from PyPI:
 pip install apache-beam
 ```
 
+#### Extra Requirements
+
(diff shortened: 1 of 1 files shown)


src/get-started/quickstart-py.md
```diff
@@ -68,6 +68,24 @@ Install the latest Python SDK from PyPI:
 pip install apache-beam
 ```
 
+#### Extra Requirements
```
aaltay: Do you think there is a need for explaining installing multiple extra packages? (I am not sure, maybe not.)

sb2nov: hmm I thought the line below should address that.

@hadarhg do you have thoughts on how to convey that the user might need to install extra packages highlighted below if they need to use certain features ?

aaltay: How about something like this at the end?

(It is possible to use multiple features by `pip install apache-beam[feature1,feature2,...]`.)

sb2nov: Done

aaltay:
LGTM

PR #213, opened 2017-04-18
[BEAM-1741] Update direct and Cloud Dataflow runner pages with Python info

melap:
R: @aaltay

### DIFF
--- a/src/documentation/runners/dataflow.md
+++ b/src/documentation/runners/dataflow.md
@@ -10,7 +10,7 @@ redirect_from: /learn/runners/dataflow/
   <strong>Adapt for:</strong>
   <ul>
     <li data-type="language-java" class="active">Java SDK</li>
-    <li data-type="language-python">Python SDK</li>
+    <li data-type="language-py">Python SDK</li>
   </ul>
 </nav>
 
@@ -58,7 +58,7 @@ For more information, see the *Before you begin* section of the [Cloud Dataflow 
 </dependency>
 ```
 
-<span class="language-python">This section is not applicable to the Beam SDK for Python.</span>
+<span class="language-py">This section is not applicable to the Beam SDK for Python.</span>
 
 ### Authentication
 
@@ -71,7 +71,7 @@ gcloud auth application-default login
 ## Pipeline options for the Cloud Dataflow Runner
 
 <span class="language-java">When executing your pipeline with the Cloud Dataflow Runner (Java), consider these common pipeline options.</span>
-<span class="language-python">When executing your pipeline with the Cloud Dataflow Runner (Python), consider these common pipeline options.</span>
+<span class="language-py">When executing your pipeline with the Cloud Dataflow Runner (Python), consider these common pipeline options.</span>
 
 <table class="table table-bordered">
 <tr>
@@ -102,11 +102,11 @@ gcloud auth application-default login
 <tr>
   <td>
     <span class="language-java"><code>tempLocation</code></span>
-    <span class="language-python"><code>temp_location</code></span>
+    <span class="language-py"><code>temp_location</code></span>
   </td>
   <td>
     <span class="language-java">Optional.</span>
-    <span class="language-python">Required.</span>
+    <span class="language-py">Required.</span>
     Path for temporary files. Must be a valid Google Cloud Storage URL that begins with <code>gs://</code>.
     <span class="language-java">If set, <code>tempLocation</code> is used as the default value for <code>gcpTempLocation</code>.</span>
   </td>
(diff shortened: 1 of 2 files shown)


src/documentation/runners/dataflow.md
```diff
@@ -51,6 +58,8 @@ You must specify your dependency on the Cloud Dataflow Runner.
 </dependency>
 ```
 
+<span class="language-python">When using Python, you do not need to specify your dependency on the Cloud Dataflow Runner.</span>
```
aaltay: Do we need this? It is a sentence about doing nothing.

src/documentation/runners/dataflow.md
```diff
@@ -61,44 +70,83 @@ gcloud auth application-default login
...
+<tr class="language-python">
+  <td><code>save_main_session</code></td>
+  <td>Save the main session state so that pickled functions and classes defined in <code>__main__</code> (e.g. interactive session) can be unpickled. Some workflows do not need the session state if, for instance, all of their functions/classes are defined in proper modules (not <code>__main__</code>) and the modules are importable in the worker.</td>
+  <td><code>false</code></td>
+</tr>
+
+<!-- Only show for Python -->
+<tr class="language-python">
+  <td><code>sdk_location</code></td>
+  <td>Override the default GitHub location from where the Cloud Dataflow SDK is downloaded. This value can be an URL, a Cloud Storage path, or a local path to an SDK tarball. Workflow submissions will download or copy the SDK tarball from this location. If set to the string <code>default</code>, a standard SDK location is used. If empty, no SDK is copied.</td>
```
aaltay: This description needs to be fixed in pipeline options as well. Could you change this to:
```
Override the default location from where the Beam SDK is downloaded. This value can be an URL, a Cloud Storage path, or a local path to an SDK tarball. Workflow submissions will download or copy the SDK tarball from this location. If set to the string <code>default</code>, a standard SDK location is used. If empty, no SDK is copied.
```

src/documentation/runners/dataflow.md
```diff
@@ -111,8 +159,11 @@ While your pipeline executes, you can monitor the job's progress, view details o
...
+To connect to your job and block until it is completed, call <span class="language-java"><code>waitToFinish</code></span><span class="language-python"><code>wait_until_finish</code></span> on the `PipelineResult` returned from `pipeline.run()`. The Cloud Dataflow Runner prints job status updates and console messages while it waits. While the result is connected to the active job, note that pressing **Ctrl+C** from the command line does not cancel your job. To cancel the job, you can use the [Dataflow Monitoring Interface](https://cloud.google.com/dataflow/pipelines/dataflow-monitoring-intf) or the [Dataflow Command-line Interface](https://cloud.google.com/dataflow/pipelines/dataflow-command-line-intf).
```
aaltay: What is the meaning of `to connect` here?

src/documentation/runners/direct.md
```diff
@@ -33,9 +47,11 @@ You must specify your dependency on the Direct Runner.
...
 ```
 
+<span class="language-python">When using Python, you do not need to specify your dependency on the Direct Runner.</span>
+
 ## Pipeline options for the Direct Runner
 
-When executing your pipeline from the command-line, set `runner` to `direct`. The default values for the other pipeline options are generally sufficient.
+When executing your pipeline from the command-line, set `runner` to `direct` or `DirectRunner`. The default values for the other pipeline options are generally sufficient.
 
 See the reference documentation for the  <span class="language-java">[`DirectOptions`]({{ site.baseurl }}/documentation/sdks/javadoc/{{ site.release_latest }}/index.html?org/apache/beam/runners/direct/DirectOptions.html)</span><span class="language-python">[`PipelineOptions`](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/utils/pipeline_options.py)</span> interface (and its subinterfaces) for defaults and the complete list of pipeline configuration options.
```
aaltay: There is also `DirectOptions`, that is probably the better link here. https://beam.apache.org/documentation/sdks/pydoc/0.6.0/apache_beam.utils.html#apache_beam.utils.pipeline_options.DirectOptions

(Assuming that we are talking about direct runner specific options.)

src/documentation/runners/dataflow.md
melap: changed to a more generic "not applicable" sentence

src/documentation/runners/dataflow.md
melap: done

src/documentation/runners/dataflow.md
melap: changed to "To block until your job completes, ..."

src/documentation/runners/direct.md
melap: done

melap:
Thanks for your feedback! I made some changes, PTAL

src/documentation/runners/dataflow.md
```diff
@@ -61,47 +70,89 @@ gcloud auth application-default login
...
+<tr class="language-python">
+  <td><code>save_main_session</code></td>
+  <td>Save the main session state so that pickled functions and classes defined in <code>__main__</code> (e.g. interactive session) can be unpickled. Some workflows do not need the session state if, for instance, all of their functions/classes are defined in proper modules (not <code>__main__</code>) and the modules are importable in the worker.</td>
+  <td><code>false</code></td>
+</tr>
+
+<!-- Only show for Python -->
+<tr class="language-python">
+  <td><code>sdk_location</code></td>
+  <td>Override the default location from where the Cloud Dataflow SDK is downloaded. This value can be an URL, a Cloud Storage path, or a local path to an SDK tarball. Workflow submissions will download or copy the SDK tarball from this location. If set to the string <code>default</code>, a standard SDK location is used. If empty, no SDK is copied.</td>
```
aaltay: please use Beam SDK instead of Cloud Dataflow SDK

melap: ACK! good catch, not sure how I managed to revert that. FIXED!

aaltay:
LGTM

melap:
SMall update to this en route, been doing battle with the javascript switcher and need to update all runner pages to specifically use language-py.

PR #216, opened 2017-04-18
Update flink runner page to use language-py instead of language-python

melap:
Hi @aljoscha I've been attempting to get the Java/Python toggling and content working properly across the docs, and noticed this page is using language-python instead of language-py, so it wasn't being toggled in tandem with the other pages/programming guide (which use language-py). This PR contains some consistency changes which might be useful, thoughts?
- change language-python to language-py
- add language toggle at the top (user currently must go to another page with a toggle to switch the page)
- language-specific change to the dependencies section to improve what's shown if set to Python (please verify if correct for Flink)
R: @aljoscha

### DIFF
--- a/src/documentation/runners/flink.md
+++ b/src/documentation/runners/flink.md
@@ -6,6 +6,14 @@ redirect_from: /learn/runners/flink/
 ---
 # Using the Apache Flink Runner
 
+<nav class="language-switcher">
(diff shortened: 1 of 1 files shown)


melap:
retest this please

aljoscha:
Thanks for fixing, @melap! 👌

PR #98, opened 2016-12-01
[BEAM-825] Fill in the documentation/runners/apex portion of the website

sandeepdeshmukh:
@tweise : Could you please review.

PR #219, opened 2017-04-21
[BEAM-825] Fill in the documentation/runners/apex portion of the website

tweise:
R: @davorbonaci 

This is on top of #98 

Please note that the example command doesn't actually use HDFS yet, the provider isn't present.


tweise:
@davorbonaci I created BEAM-2053 - please comment if the suggested change to the examples makes sense.

davorbonaci:
(makes sense)

PR #229, opened 2017-05-03
Transfer some content from Create Your Pipeline to Programming Guide

hadarhg:


hadarhg:
R: @davorbonaci (This is instead of https://github.com/apache/beam-site/pull/221 - I closed that one)

PR #232, opened 2017-05-08
[BEAM-2222] Move content from apex readme to website

aaltay:
After this `runners/apex/README.md` can be removed.

R: @davorbonaci


davorbonaci:
This is not really necessary; there's nothing specific to the apex runner here. I'd just drop it.

aaltay:
Makes sense.

PR #236, opened 2017-05-09
[BEAM-2222] fix typos in hadoop documentation

aaltay:
R: @hadarhg

### DIFF
--- a/src/documentation/io/built-in-hadoop.md
+++ b/src/documentation/io/built-in-hadoop.md
@@ -70,7 +70,7 @@ p.apply("read",
 
 #### Read data with configuration and key translation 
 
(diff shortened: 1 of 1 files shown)


hadarhg:
- [ ] Change:
For example scenario: Beam Coder is not available for key class hence key translation is required.
To:
For example, a Beam Coder is not available for the `Key` class, so key translation is required.

- [ ] Change: 
For example scenario: Beam Coders are not available for both `Key` class and `Value` class of `InputFormat` hence key and value translation is required.
To:
For example, Beam Coders are not available for the `Key` and `Value` classes of `InputFormat`, so key and value translation are required.

aaltay:
Thank you @hadarhg. PTAL.

PR #242, opened 2017-05-12
[BEAM-1963] Quick start should not default to java quickstart

aaltay:
@melap What do you think? Would it be better to have a single quick start index page with links to each language?


davorbonaci:
I think this is a good change in the current world (via capitalized names: Java / Python). Going forward, I'm hoping all documentation pages will have toggles, so it can default to whatever the toggle defaults to.

aaltay:
Capitalized the language names, PTAL.

src/get-started/index.md
```diff
@@ -14,7 +14,7 @@ Learn to use Beam to create data processing pipelines that run on supported proc
 
 Learn about the Beam model, the currently available Beam SDKs and Runners, and Beam's native I/O connectors.
 
-#### [Quickstart]({{ site.baseurl }}/get-started/quickstart/)
+#### Quickstart [Java]({{ site.baseurl }}/get-started/quickstart-java) or [Python]({{ site.baseurl }}/get-started/quickstart-py)
```
melap: For all of these, I'd suggest adding "for": Quickstart for [Java] or [Python]

aaltay:
@melap addressed your comment, PTAL.

PR #246, opened 2017-05-17
Update Java Quickstart to stop using snapshot repos

dhalperi:
And always use the latest Beam release by name.


dhalperi:
R: @davorbonaci

PR #249, opened 2017-05-23
[BEAM-1158]HCatalog example using HadoopIputFormatIO

seshadri-cr:
This is an example of reading through HCatalog using HadoopInputFormatIO. Reference JIRA - https://issues.apache.org/jira/browse/BEAM-1158
Please let me know if any further details are required.

### DIFF
--- a/src/documentation/io/built-in-hadoop.md
+++ b/src/documentation/io/built-in-hadoop.md
@@ -217,26 +217,10 @@ org.apache.hive.hcatalog.mapreduce.HCatInputFormat.setInput(hcatConf, "my_databa
 Call Read transform as follows:
 
 ```java
-PCollection<KV<Long, String>> hcatData =
+PCollection<KV<Long, HCatRecord>> hcatData =
   p.apply("read",
-  HadoopInputFormatIO.<Long, String>read()
-  .withConfiguration(hcatConf)
-  .withValueTranslation(hcatOutputValueType);
-```
-
-```py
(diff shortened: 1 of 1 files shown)


aaltay:
R: @ssisk

src/documentation/io/built-in-hadoop.md
```diff
@@ -195,3 +195,50 @@ PCollection<KV<Text, LinkedMapWritable>> elasticData = p.apply("read",
...
+```
+
+Call Read transform as follows:
+
+```java
+PCollection<KV<Long, String>> hcatData =
+  p.apply("read",
+  HadoopInputFormatIO.<Long, String>read()
+  .withConfiguration(hcatConf)
+  .withValueTranslation(hcatOutputValueType);
```
iemejia: I think you can remove this withValueTranslation and the function because the most important part for a user is to get the PCollection of HCatRecord, then it is up to him to decide if he maps it to strings or to his own object type.

seshadri-cr: makes sense, removed the valueTranslation part in the latest commit !

seshadri-cr:
@iemejia , since we have reached a consensus that HIFIO examples for datastores that do not have a native IO implementation should go to the beam-site hope we can go ahead and merge this PR(if there are no more review comments) and subsequently close the JIRA - https://issues.apache.org/jira/browse/BEAM-1158, WDYT.

iemejia:
LGTM too, merging

PR #254, opened 2017-05-26
[BEAM-1073] Do not use BeautifulSoup prettify, which is known to corrupt HTML

kennknowles:
Opening this to see what happens to the staged HTML when we don't prettify.


kennknowles:
R: @dhalperi

dhalperi:
LGTM, nice.

dhalperi:
Merged, thanks. Not sure why this was previously believed hard to fix :/

davorbonaci:
(just saw this, thank you!)

CC: @jasonkuster

PR #262, opened 2017-06-27
Update Flink Runner doc

aljoscha:
This updates to a more recent Flink version and also adds a note about matching Scala versions.

R: @jingsongli

### DIFF
--- a/src/documentation/runners/flink.md
+++ b/src/documentation/runners/flink.md
@@ -40,7 +40,7 @@ $ mvn dependency:tree -Pflink-runner |grep flink
 [INFO] |  +- org.apache.flink:flink-streaming-java_2.10:jar:1.3.0:runtime
 ...
 ```
-Here, we would need Flink 1.3.0.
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/beam_PreCommit_Website_Stage/541/

Jenkins built the site at commit id c4aa8d98b129414cce22f6c3db0123342ec3123b with Jekyll and staged it [here](http://apache-beam-website-pull-requests.storage.googleapis.com/262/index.html). Happy reviewing.

JingsongLi:
LGTM.

ChristophHebert:
@aljosch The clarification about the Scala version is good.

Note that the word-count-beam example acquired from the Quickstart - Java guide currently produces the following result from `mvn dependency:tree` :
`[INFO]    +- org.apache.flink:flink-streaming-java_2.10:jar:1.2.1:runtime`
So it might be premature to update the Flink version to 1.3.0 as done in this PR. (Unless the example acquired from the Quickstart guide is expecting an update soon).

aljoscha:
@ChristophHebert You're right that it's not the current version, it will be the Flink version that the next Beam release (2.1.0) will use. I'll change it to use the version that we currently have.

@JingsongLi thanks for reviewing!

PR #274, opened 2017-07-14
[BEAM-1669] IO testing - unit test docs

ssisk:
hey,

I've ported the content from the "Testing I/O Transforms in Apache Beam" doc over to the website. I've only done the unit test section - integration testing comes next.

PTAL

cc @melap @jkff

### DIFF
--- a/src/documentation/io/testing.md
+++ b/src/documentation/io/testing.md
@@ -99,7 +99,7 @@ Instead of using mocks in your unit tests (pre-programming exact responses to ea
 
 ### Network failure
 
-To help with testing and separation of concerns, **code that interacts across a network should be handled in a separate class from your I/O transform**. The suggested design pattern is that your I/O transform throws exceptions once it determines that a read is no longer possible.
+To help with testing and separation of concerns, **code that interacts across a network should be handled in a separate class from your I/O transform**. The suggested design pattern is that your I/O transform throws exceptions once it determines that a read or write is no longer possible.
 
 This allows the I/O transform's unit tests to act as if they have a perfect network connection, and they do not need to retry/otherwise handle network connection problems.
 


asfgit:
https://builds.apache.org/job/beam_PreCommit_Website_Stage/570/

Jenkins built the site at commit id b14837ca33702b4e4d41fdfd6c0e2d728300c1a9 with Jekyll and staged it [here](http://apache-beam-website-pull-requests.storage.googleapis.com/274/index.html). Happy reviewing.

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
...
+Our test strategy is a balance of those 2 contradictory needs. We recommend doing as much testing as possible in unit tests, and writing a single, small integration test that can be run in various configurations.
+
+
+## Examples {#examples}
+
+
+
+*   `BigtableIO`'s testing implementation is considered the best example of current best practices for unit testing `Source`s. 
+*   `DatastoreIO` best demonstrates usage of the Service interface design pattern.
+*   `JdbcIO` has the current best practice examples for writing integration tests.
```
jkff: Would be good to link to code on github. Also, I think we have more interesting examples of unit tests - e.g. ElasticsearchIO is a very good example for bounded read/write, and MqttIO and AmqpIO are good examples for unbounded.

By the way, testing streaming sources/sinks is particularly challenging, so we definitely must reference them, and specifically call out the testing of streaming sources/sinks as a section in this document.

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
...
+## Unit Tests {#unit-tests}
+
+
+### Goals {#goals}
+
+
+
+*   Validate the correctness of the code in your I/O transform.
+*   Validate that the I/O transform works correctly when used in concert with reference implementations of the data store it connects with (where "reference implementation" means a fake or in-memory version).
+*   Be able to run quickly (< 1 sec) and need only one machine, with a reasonably small memory/disk footprint and no non-local network access (preferably none at all).
```
jkff: 1 second is often an unrealistically strict requirement that even our own unit tests don't satisfy - especially e.g. the exhaustive source tests. I'd suggest to shoot for a couple of seconds if possible; with a hard limit of, say, 20 seconds. E.g. a testing instance of Hive takes several seconds to bring up and shut down and the full unit test takes about 20s; it's not bad enough to make us have only integration tests for it.

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
...
+*   Test problems in the external data store - this can lead to extremely complicated tests.  
+
+
+### Implementing unit tests {#implementing-unit-tests}
+
+A general guide to writing Unit Tests for all transforms can be found in the [PTransform Style Guide](https://beam.apache.org/contribute/ptransform-style-guide/#testing ). We have expanded on a few important points below.
+
+If you are implementing a `Source`/`Reader` class, make sure to exhaustively unit-test your code. A minor implementation error can lead to data corruption or data loss (such as skipping or duplicating records) that can be hard for your users to detect. Also look into using `SourceTestUtils` - it is a key piece of test `Source` implementations.
+
+If you are not using the `Source` API, you can use DoFnTester to help with your testing. Datastore's I/O transforms have some good examples of how to use it in testing I/O transforms.
```
jkff: I think in the majority of cases, if you're not using the Source API, you should use TestPipeline and PAssert, rather than DoFnTester.

Also note that PAssert is only useful for testing the read part of an IO transform. To test the write part, you need to run a TestPipeline that writes the data and then inspect the storage instance under test. Probably also worth calling out explicitly in the document, that testing reads/writes is different.

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
...
+
+
+### Use fakes {#use-fakes}
+
+Instead of using mocks in your unit tests (pre-programming exact responses to each call for each test), use fakes (a lightweight implementation of the service that behaves the same way at a very small scale) or an in-memory version of the service you're testing. This has proven to be the right mix of "you can get the conditions for testing you need" and "you don't have to write a million exacting mock function calls".
```
jkff: I'd give stricter guidance here: if there exists a lightweight embeddable version of the service you're testing, then absolutely you should use that.

Only if it's impossible or unusable (e.g. Cassandra requires Java8) - then you should develop a fake, or hide it under an interface. Good examples are CassandraIOTest (with a pretty lightweight fake) and BigQueryIO tests (with a very involved fake, but still better than talking to the real BigQuery).

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
...
+
+
+### Network failure
+
+To help with testing and separation of concerns, **code that interacts across a network should be handled in a separate class from your I/O transform**, and thus should be unit tested separately from the I/O transform itself. In many cases, necessary network retries should be encapsulated within a fake implementation.
```
jkff: Do we have any examples of such testing?

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
...
+### Network failure
+
+To help with testing and separation of concerns, **code that interacts across a network should be handled in a separate class from your I/O transform**, and thus should be unit tested separately from the I/O transform itself. In many cases, necessary network retries should be encapsulated within a fake implementation. 
+
+The suggested design pattern is that your I/O transform throws exceptions once it determines that a read is no longer possible. This allows the I/O transform's unit tests to act as if they have a perfect network connection, and they do not need to retry/otherwise handle network connection problems.
+
+
+## Batching
+
+If your I/O transform allows batching of reads/writes, you must force the batching to occur in your test. Having configurable batch size options on your I/O transform allows that to happen easily (potentially marked as test-only)
```
jkff: It should basically always be marked as test-only. In all cases where a connector allowed to configure a batch size, we reached consensus that it should not be exposed to users, but is useful in tests.

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
 > Note: This guide is still in progress. There is an open issue to finish the guide: [BEAM-1025](https://issues.apache.org/jira/browse/BEAM-1025).
 
 
+## Testing IO Transforms in Apache Beam
```
melap: IO  ->  I/O  (to be consistent with everything else)

src/documentation/io/io-toc.md
```diff
@@ -17,9 +17,11 @@ permalink: /documentation/io/io-toc/
 
 * [Authoring I/O Transforms - Overview]({{site.baseurl }}/documentation/io/authoring-overview/)
 
+* [Testing I/O Transforms]({{site.baseurl }}/documentation/io/testing/)
+
 <!-- TODO: commented out until this content is ready.
 * [Authoring I/O Transforms - Python]({{site.baseurl }}/documentation/io/authoring-python/)
 * [Authoring I/O Transforms - Java]({{site.baseurl }}/documentation/io/authoring-java/)
-* [Testing I/O Transforms]({{site.baseurl }}/documentation/io/testing/)
+
```
melap: extra blank line creates separation in list bullets

src/documentation/io/testing.md
```diff
@@ -11,9 +11,102 @@ permalink: /documentation/io/testing/
...
+    *   Data set size: tiny (10s to 100s of rows)
+*   **Integration Tests:**
+    *   Goal: catch problems that occur when interacting with real versions of the runners/data store
+    *   Data store used: an actual instance, pre-configured before the test
+    *   Data set size: small to medium (1000 rows to 10s of GBs)
+
+
+## A note on performance benchmarking
+
+We do not advocate writing a separate test specifically for performance benchmarking. Instead, we advocate setting up integration tests so that they can be parameterized in a way that allows for covering many different testing scenarios.
```
melap: hmm... trying to pare that second sentence down a bit... is this (or something similar) still true to your intention?
Instead, we recommend setting up integration tests that can accept the necessary parameters to cover many different testing scenarios.

src/documentation/io/testing.md
melap: is that "of" in "testing and separation of concerns" extra?

src/documentation/io/testing.md
melap: DoFnTester not in code font. but this brings up another question -- does this guide also apply to Python or is it Java only?

src/documentation/io/testing.md
ssisk: done

src/documentation/io/testing.md
ssisk: I can't think of any off the top of my head. Thinking about this more - what I've generally seen has been that the "code that interacts across a network" ends up being the client. I've adapted this paragraph to fit that.

ssisk: https://en.wikipedia.org/wiki/Separation_of_concerns is what I'm referencing here. So the 'of' is not extra

src/documentation/io/testing.md
ssisk: done

src/documentation/io/testing.md
```diff
@@ -6,14 +6,113 @@ permalink: /documentation/io/testing/
...
+
+
+### Use fakes {#use-fakes}
+
+Instead of using mocks in your unit tests (pre-programming exact responses to each call for each test), use fakes. The preferred way to use fakes for I/O transform testing is to use a pre-existing in-memory/embeddable version of the service you're testing, but if one does not exist consider implementing your own. Fakes have proven to be the right mix of "you can get the conditions for testing you need" and "you don't have to write a million exacting mock function calls".
+
+
+### Network failure
+
+To help with testing and separation of concerns, **code that interacts across a network should be handled in a separate class from your I/O transform**. The suggested design pattern is that your I/O transform throws exceptions once it determines that a read is no longer possible.
```
jkff: a read or write

src/documentation/io/testing.md
```diff
@@ -6,14 +6,113 @@ permalink: /documentation/io/testing/
...
+
+This allows the I/O transform's unit tests to act as if they have a perfect network connection, and they do not need to retry/otherwise handle network connection problems.
+
+
+## Batching
+
+If your I/O transform allows batching of reads/writes, you must force the batching to occur in your test. Having configurable batch size options on your I/O transform allows that to happen easily. These must be marked as test only.
 
+<!--
```
jkff: Does the current PR intentionally comment this out?

src/documentation/io/testing.md
ssisk: done

src/documentation/io/testing.md
ssisk: yes. The "contributing" doc has not been added yet, so we should not be linking to it.

Before this CL, /testing/ (this page) was originally present but unreachable from any other pages on the site. In this CL, in io-to, we added a link to /testing/ (this page), but want to make sure we don't surface the link to the contributing doc until it's ready.

asfgit:
https://builds.apache.org/job/beam_PreCommit_Website_Stage/579/

Jenkins built the site at commit id a8ed68aee4a31c52f0b5be20e79890de66017fd9 with Jekyll and staged it [here](http://apache-beam-website-pull-requests.storage.googleapis.com/274/index.html). Happy reviewing.

jkff:
@asfgit merge

asfgit:
Error: User jkff not a committer on beam-website or not registered in gitbox.apache.org; access denied.

jkff:
ok I've initiated some account-related process and I'll try this command again in an hour or so.

jkff:
@asfgit merge

PR #308, opened 2017-08-29
[MINOR] scrollable dropdown menu

vectorijk:
support for scrollable dropdown menu

image

### DIFF
--- a/content/css/site.css
+++ b/content/css/site.css
@@ -3201,7 +3201,9 @@ tbody.collapse.in {
   border-radius: 4px;
   -webkit-box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175);
   box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175);
-  background-clip: padding-box; }
+  background-clip: padding-box;
+  max-height: 400px;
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/beam_PreCommit_Website_Stage/662/

Jenkins built the site at commit id ce98c289953001f44172c6b088528143ae229a4a with Jekyll and staged it [here](http://apache-beam-website-pull-requests.storage.googleapis.com/308/index.html). Happy reviewing.

reuvenlax:
R: @melap

vectorijk:
@melap Since you will improving the navigation story, should I close this PR?
Eventually, I think we should support scrollable pulldown menu. It is too long to see all items.

melap:
no, I agree, I think it's very useful and fixes a problem for small windows until future improvements are finished. my only concern is that it does make things less discoverable if looking at the menu in a full-size browser window , as all items are not visible at once anymore. so I was wondering if there's a way to only do the scrollbar if the window is smaller than the menu pulldown. maybe there is no way, I am not sure.

vectorijk:
#332 solved. I would close this.

PR #336, opened 2017-10-19
[BEAM-664] Update Java MinimalWordCount instructions

alex-filatov:
MinimalWordCount in Java [is intentionally hardcoded](https://github.com/apache/beam/pull/1315) to run only on DirectRunner.

### DIFF
--- a/src/get-started/wordcount-example.md
+++ b/src/get-started/wordcount-example.md
@@ -50,44 +50,8 @@ input and output sources and show other best practices.
 
 **To run this example in Java:**
 
-{:.runner-direct}
 ```
 $ mvn compile exec:java -Dexec.mainClass=org.apache.beam.examples.MinimalWordCount
-```
-
-{:.runner-apex}
-```
-$ mvn compile exec:java -Dexec.mainClass=org.apache.beam.examples.MinimalWordCount \
-     -Dexec.args="--inputFile=pom.xml --output=counts --runner=ApexRunner" -Papex-runner
-```
-
-{:.runner-flink-local}
-```
-$ mvn compile exec:java -Dexec.mainClass=org.apache.beam.examples.MinimalWordCount \
-     -Dexec.args="--runner=FlinkRunner --inputFile=pom.xml --output=counts" -Pflink-runner
-```
-
-{:.runner-flink-cluster}
-```
-$ mvn package exec:java -Dexec.mainClass=org.apache.beam.examples.MinimalWordCount \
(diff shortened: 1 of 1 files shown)


aaltay:
Hey @alex-filatov, could you explain this more? Should we instead change the example to not-hardcode to work only one runner.

cc: @kennknowles in case he has more context.

alex-filatov:
@aaltay @kennknowles Current instructions are broken: you'll get exception if you try to run Java MinimalWordCount using e.g. Flink runner. Checking history, it appears this was a deliberate change. Commit message: "This makes it easy to immediately run, and removes various
non-portable instructions and others that aren't the easiest for a "Getting Started" scenario." This makes sense to me. 

> Should we instead change the example to not-hardcode to work only one runner.

Not sure. Beam model is independent of the runners and could be explained separately (using Direct Runner) from the real runners. I would change _all_ examples to use Direct Runner only and provide separate instructions how to specify and configure different runners. IMO this decoupling would make docs simpler.

kennknowles:
The context is that just to keep it truly "minimal" as a demonstration of how pithy we could get things with Java 8, we hardcode everything. The idea is that if you want to experiment with it you edit the code. It is just the tiniest example.

kennknowles:
That was back in a time where runners needed more special consideration than they do now. As long as the code for `MinimalWordCount` stays very concise, I am happy to have command lines that run on different runners.

aaltay:
@kennknowles My understanding so far is, {{MinimalWordCount}} needs pipeline options for runner, input, output to be able to run with other runners. We do not want to do that to complicate things. In that case, making this documentation changes makes sense. Is this accurate?

kennknowles:
Actually it has [harcoded input](https://github.com/apache/beam/blob/master/examples/java/src/main/java/org/apache/beam/examples/MinimalWordCount.java#L78) and [hardcoded output](https://github.com/apache/beam/blob/master/examples/java/src/main/java/org/apache/beam/examples/MinimalWordCount.java#L114). I do think that you could set the runner via pipeline options and it would work.

aaltay:
Does the hardcoded input/output work for all runners as it is?

kennknowles:
I'm actually not sure. FWIW this PR is clearly an improvement as the bits it deletes are passing command line args that don't exist. I expect they would actually cause the command to fail.

aaltay:
I agree that this PR is an improvement. I can merge it unless there is an objection.

cc: @melap

melap:
LGTM

PR #339, opened 2017-10-31
[BEAM-3121] Remove broken docker script and documentation

herohde:


herohde:
Hmm. This PR doesn't actually change the website, so I'm surprised the checks failed.

PR #365, opened 2017-12-14
Update the release guide after 2.2.0 release experience

kennknowles:


kennknowles:
R: @reuvenlax I cannot add you as a reviewer to this PR, which seems odd. I would assume that is part of gitbox sync, since I can add everyone else. Perhaps you are bitten by the bug where many people were removed from the org https://github.com/apache?

melap:
@asfgit merge

PR #383, opened 2018-01-31
BEAM-3575 - Update contribution guidelines to add the option of cloni…

coheigea:
…ng from a fork


coheigea:
OK thanks for the feedback - I had manually forked Beam and then got a bit confused by the instructions on the webpage, hence the PR. If you feel that the instructions are clear enough "as is" then I'm OK with closing the PR.

PR #389, opened 2018-02-16
Navigation side bar expand and collapse

skarensmoll:
There was included the needed behaviour on navigation side bar so as to collapse and expand menu.

### DIFF
--- a/content/css/site.css
+++ b/content/css/site.css
@@ -6560,24 +6560,21 @@ body {
 .section-nav {
   max-width: 250px;
   position: fixed;
-  width: 100%; }
+  width: 100%;
+  overflow-y: auto; }
   .section-nav nav {
     background-color: #F7F7F7;
-    padding: 15px;
-    max-height: calc(100vh - 130px);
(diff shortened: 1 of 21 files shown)


melap:
Thanks for the update and fixes! I checked it out on staging and it looks like the left nav now sometimes cuts off the items and has a scrollbar, even in a large window that can display everything. Has the calculation of when to show the scrollbar changed? Before, it showed all items unless the browser window was too small to show them and only then made a scrollbar. I noticed it on the Contribute section, you can go back and forth between these two links to see the before and after:

Current site (can see all items on a large screen)
https://beam.apache.org/contribute/

With these changes (scrollbar, can't see all items)
http://apache-beam-website-pull-requests.storage.googleapis.com/389/contribute/index.html

skarensmoll:
Hi Melissa, I'm seeing some changes of mine were not included on the pull request for some reason...
I'm gonna solve that right away thanks for your comments you're completely right !

melap:
Closing this PR, and keeping #403

PR #391, opened 2018-02-21
Fixed number of contributors and duplication of name of Alexey Romanenko

aromanenko-dev:


iemejia:
Thanks that was my mistake trying to change your previous name aromanenko-dev for Alexey Romanenko.

iemejia:
@asfgit merge

PR #412, opened 2018-04-06
[BEAM-3985] Document Gradle commands for testing and linting.

udim:
Also fixes outdated references to lint_py2 and lint_py3.

PR #414, opened 2018-04-06
[BEAM-3252] Updating contributor's guide to reference Gradle.

youngoli:
Removed as many references to maven or pom files under contribute
as I could. There's still a few references in the Eclipse Tips file
since I have no clue how to use Eclipse with Gradle and have yet to
find someone who does.

Also references in maturity-model in the dependency tree. It's not
really possible to create a Gradle dependency tree that isn't hundreds
of thousands of lines though, because each distinct configuration in
each project has it's own dependencies so you end up creating hundreds
of dependency trees, many of which are hundreds of lines long by
themselves.

### DIFF
--- a/src/contribute/intellij.md
+++ b/src/contribute/intellij.md
@@ -41,7 +41,7 @@ IntelliJ supports checkstyle within the IDE using the Checkstyle-IDEA plugin.
 
 1. Install the "Checkstyle-IDEA" plugin from the IntelliJ plugin repository
 2. Configure the plugin by going to Settings -> Other Settings -> Checkstyle
-3. Set Checkstyle version to the same as in `/build_rules.gradle` (e.g. 8.2)
+3. Set Checkstyle version to the same as in `/build_rules.gradle` (e.g. 8.7)
 4. Set the "Scan Scope" to "Only Java sources (including tests)"
 5. In the "Configuration File" pane, add a new configuration using the plus icon:
     1. Set the "Description" to "Beam"


youngoli:
Note: I was making this PR unaware that @udim was working on an overlapping change in #412. The overlapping changes will probably need to be merged by someone.

Also note: This change and other Maven->Gradle changes should probably go in a separate branch so the site doesn't mention Maven in some places and Gradle in others.

src/contribute/intellij.md
```diff
@@ -15,6 +15,22 @@ section_menu: section-menu/contribute.html
...
+3. Under Project Structure > Modules, click the + sign to add a module and select "Import Module".
+    1. Select the directory containing the Beam source tree.
+    2. Tick the "Import module from external model" button and select Gradle from the list.
+    3. Tick the following boxes.
+       * Use auto-import
+       * Create separate module per source set
+       * Store generated project files externally
+       * Use default gradle wrapper
+
+This should result in a working Gradle project. Build the project by executing the "build" task in
```
lukecwik: I believe we should tell users to delegate build/test actions to gradle. We should tell users to do this as well, instructions are here: https://dzone.com/articles/using-idea-to-delegate-build-and-run-actions-to-gr

src/contribute/intellij.md
```diff
@@ -15,6 +15,22 @@ section_menu: section-menu/contribute.html
...
+the root Gradle module.
+
 ## Enable Annotation Processing
```
lukecwik: I don't believe we need this whole section.

lukecwik:
@aaltay / @udim For the Python SDK development in the contribution guide, should we prefer gradle since it automatically sets up the virtualenv. It seems like a lot less work for new contributors. There can be Python SDK advanced instructions, similar to how the intellij instructions exists which go through the virtualenv setup, ...

aaltay:
My opinion is that, it is better to use gradle based python development in the contribution guide. It makes it uniform across the project and easy to get started. I would think that anyone doing heavy development on python will actually use virtualenv based setup. Let's keep them as an alternative (or advanced) option.

src/contribute/intellij.md
youngoli: By this whole section do you mean the "Enable Annotation Processing" section?

lukecwik: Yes, `target` folders refer to Maven.

Also, try out the quickstart manually using intellij in a new project.

youngoli: Ah. So does that mean there's no gradle version of those directories then if the solution is to remove the entire section?

Also yes, I'll do that. I already did so for the gradle setup directions, but I ignored the rest of the instructions (annotations, checkstyle, and codestyle).

lukecwik: Based upon your current set of instructions for intellij, I don't believe we need anything additional for the annotation processing.

lukecwik:
@asfgit merge

PR #417, opened 2018-04-11
Change first occurance of MyOptions to PipelineOptions

coheigea:
In the programming guide, section 2.1. Configuring pipeline options, "MyOptions" is used as the variable name when showing how to set PipelineOptions from command-line arguments. However this is a bit confusing, as MyOptions is only explained in the following section "2.1.2. Creating custom options". It's better instead just to use PipelineOptions here.


melap:
@asfgit merge

PR #418, opened 2018-04-11
[BEAM-3995] Update Nexmark launch instructions for Gradle

kennknowles:
This ports the instructions to Gradle. There are a couple of known issues, but I thought a PR updating the docs was a great place to make sure the new invocations work.

### DIFF
--- a/src/documentation/sdks/nexmark.md
+++ b/src/documentation/sdks/nexmark.md
@@ -9,12 +9,14 @@ permalink: /documentation/sdks/java/nexmark/
 ## What it is
 
 Nexmark is a suite of pipelines inspired by the 'continuous data stream'
-queries in [Nexmark research paper](http://datalab.cs.pdx.edu/niagaraST/NEXMark/)
+queries in [Nexmark research
+paper](http://datalab.cs.pdx.edu/niagaraST/NEXMark/)
 
-These are multiple queries over a three entities model representing on online auction system:
+These are multiple queries over a three entities model representing on online
+auction system:
 
- - **Person** represents a person submitting an item for auction and/or making a bid
-    on an auction.
+ - **Person** represents a person submitting an item for auction and/or making
+   a bid on an auction.
  - **Auction** represents an item under auction.
  - **Bid** represents a bid for an item under auction.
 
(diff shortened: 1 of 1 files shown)


kennknowles:
@jbonofre @echauchot @akedin @apilloud

src/documentation/sdks/nexmark.md
```diff
@@ -112,17 +122,19 @@ Performance:
...
+to this, the necessary dependencies must be configured.
 
-The runner must also be specified like in any other Beam pipeline using:
+When running via Gradle, the following two parameters control the execution:
 
-    --runner
+    -P nexmark.args
+        The command line to pass to the Nexmark main program.
 
+    -P nexmark.runner
```
echauchot: no more profiles for the runners then? The needed dependencies are all included in the runner projet specified by the user?

src/documentation/sdks/nexmark.md
```diff
@@ -417,124 +429,192 @@ Yet to come
...
+            --manageResources=false
+            --monitorJobs=true
+            --enforceEncodability=true
+            --enforceImmutability=true"
 
 ### Running SMOKE suite on the SparkRunner (local)
 
+The SparkRunner is special-cased in the Nexmark gradle launch. The task will
+provide the version of Spark that the SparkRunner is built against, and
+configure logging.
```
echauchot: Sorry but I don't understand this comment: owing to what's bellow, the version is automatically provided and no parameter needs to be set by the user. So why not just remove this comment if it is not targeted to the user of nexmark?
Is there something I missed?

src/documentation/sdks/nexmark.md
kennknowles: Yes, Gradle puts them on the classpath. The one exception is the Spark runner. I used to have a `-P nexmark.provideSpark` parameter to keep it a little decoupled, but since it is pretty much still "magic" knowledge, I followed @lukecwik's suggestion and just hardcoded knowledge of the dependency, so I did that [here](https://github.com/apache/beam/blob/master/sdks/java/nexmark/build.gradle#L35).

I think when I add the fat jar capability it may require a separate fat jar for each known runner config. At that point, pretending that the runner dependency can be anything (versus a selection of a hardcoded profile) may not be useful.

src/documentation/sdks/nexmark.md
kennknowles: This is a comment about the hardcoding above. The nexmark build tweaks the dependencies based on special needs of the SparkRunner instead of being just a generic suite of pipelines. I will delete the comment here since it "just works" for users copying and pasting from here.

src/documentation/sdks/nexmark.md
kennknowles: The alternative which I mean is that we could just make it like before where you choose a profile and `-P nexmark.runner` is a choice from an enum, not a magic Gradle subproject name.

echauchot: Yes I think I prefer this explicit version rather than the magic one.

kennknowles: OK. Filed https://issues.apache.org/jira/browse/BEAM-4139 in the gradle improvements to go back to the repo and change it.

kennknowles:
@asfgit merge

asfgit:
Error: PR failed in verification; check the Jenkins job for more information.

asfgit:
Error: PR failed in verification; check the Jenkins job for more information.

melap:
retest this please

kennknowles:
@asfgit merge

asfgit:
Error: PR failed in verification; check the Jenkins job for more information.

melap:
@asfgit merge

melap:
the robot likes me right now @kennknowles  ;)

PR #422, opened 2018-04-18
Add TODO with linked JIRA to update Docker documentation for Gradle.

swegner:


swegner:
Adding a TODO so we don't forget to come back to it.

R: @melap 
CC: @iemejia

melap:
Is it worth making this visible for anyone who visits the page and sees that it isn't updated? That way they can go look at the issue status themselves, sign up to help, etc. (also it's less "out of sight, out of mind" )
```
> Note: There is an open issue to update this page for Gradle build: [BEAM-4117](https://issues.apache.org/jira/browse/BEAM-4117).
```
(example of how the > prefixed note renders at https://beam.apache.org/documentation/io/testing/)

swegner:
Love it. Updated. @melap

melap:
Thanks, merged

PR #423, opened 2018-04-19
Document Java extensions for parsing Apache HTTPD logfiles and Useragent strings

nielsbasjes:
I spoke to @davorbonaci yesterday and he thought it would be a good idea to document the existence of these two projects that work in combination with Apache Beam.

Please review if you guys want this.

I'll review the output of the CI build to check if I have to fix some things.


nielsbasjes:
test this please

nielsbasjes:
I rebased it. 
This looks good on my machine after running the Jekyll generator.

iemejia:
This sounds like a really good idea, but I don't think we should mix these with the extensions that are hosted in Apache. @melap WDYT ? Maybe we create a webpage for external extensions or 'beam friend projects' where we can also mention others like scio ?

davorbonaci:
(Either way works for me. Thanks @nielsbasjes.)

nielsbasjes:
Likewise; either way is fine by me.
Do you want me to move this to the desired location? If so, then please specify where you want it.
You guys moving it to the desired location is also fine.

melap:
sorry, am traveling so slow to respond. @iemejia I agree that this is an interesting idea, and would be good to put on its own page that has Beam-powered projects. I am wondering if it would be best to list the projects with a 2-3 sentence description and a link to the project site for each, and the full documentation is better living at each project's site? (in this case @nielsbasjes Github?) Long term I don't think having the docs for all Beam related projects on our site will scale. I'm thinking have something like this on a new page? https://flink.apache.org/ecosystem.html#third-party-projects

nielsbasjes:
I'll put up a new version based on what you guys said. I have to check what makes the most sense.

nielsbasjes:
I moved this to a separate page and rewrote the content to be more compact. 
Please check if this is what you guys want.

nielsbasjes:
http://apache-beam-website-pull-requests.storage.googleapis.com/423/documentation/sdks/java-thirdparty/index.html

melap:
Thanks @nielsbasjes ! Looks good -- as @davorbonaci suggested, I'll merge now and will ponder on how we can continue to scale this as we get more projects going forward.

melap:
@asfgit merge

PR #435, opened 2018-05-04
Fix/remove references to the old teams page

kennknowles:
@melap I removed those that seemed obsolete and fixed those that seemed to still add value. Our Jenkins links are still timing out a bit, and I excluded datatorrent.com from tests since it is wholly down.


melap:
Looks fine, only thing I note is at some point we probably want to make another external url icon that matches the text color for the footer, but not a high priority.

PR #467, opened 2018-06-12
[BEAM-4545]Update release guide with how to build python wheel

boyuanzz:
r: @robertwb @aaltay

### DIFF
--- a/src/contribute/release-guide.md
+++ b/src/contribute/release-guide.md
@@ -423,16 +423,12 @@ Verify that files are [present](https://dist.apache.org/repos/dist/dev/beam).
      ```
    * Add secure to ./travis.yml file
      ```
-     travis encrypt SOMEVAR="secretvalue" --add
+     travis encrypt -r github_suername/apache-beam-wheels WHEELHOUSE_UPLOADER_SECRET=your pypi password --add
      ```
 1. Commit changes and push into your repo.
 1. Trigger build:
    * Sign up [Travis CI](https://travis-ci.org/) with your github account.
    * Choose ```apache-beam-wheels``` to build.
-   * Setup Environment Variables:
-     * Register test.pypi account: https://test.pypi.org/account/register/ .
-     * Goto ```More options``` and choose ```Settings```.
-     * In ```Environment Variables```, add value pair <WHEELHOUSE_UPLOADER_SECRET, your test pypi account password>.
    * In ```More options```, click ```Trigger build```.
 
 


boyuanzz:
I'm not sure whether Release Manager is supposed to have himself own pypi account. If it's better to let all Release Manager use 'apache-beam-travis', I'll update. @robertwb @aaltay

src/contribute/release-guide.md
```diff
@@ -398,6 +398,40 @@ Staging binaries
 
 Verify that files are [present](https://dist.apache.org/repos/dist/dev/beam).
 
+
+### Build python wheels and upload it to test.pypi.org
+
+1. Create your test.pypi.org account in https://test.pypi.org/account/register/ .
+   * Note: Please ping dev@ thread to ask for permissions accessing apache_beam in test.pypi.org.
+1. Fork https://github.com/robertwb/apache-beam-wheels .
```
tvalentyn: Perhaps this can live in beam/release-tools or something like that?

boyuanzz: Hey Valentyn, could you please explain more about your ideas?

tvalentyn: I was thinking that perhaps it would make sense to move necessary tooling to build wheels into some folder in Beam Github repository that contains release tools?

boyuanzz: From what I know so far, it's better to keep it as a separated repo. How about you @robertwb ?

src/contribute/release-guide.md
```diff
@@ -398,6 +398,40 @@ Staging binaries
 
 Verify that files are [present](https://dist.apache.org/repos/dist/dev/beam).
 
+
+### Build python wheels and upload it to test.pypi.org
```
robertwb: We should put some TODOs in here to make it clear this is no the long-term process. 

Probably we should just have a single repo apache/beam-wheels (which all committers can push to) and upload to somewhere other than test.pypi.org.

boyuanzz: I agree add TODOs here, but I'm not sure whether it's OK to have TODO in public doc.

src/contribute/release-guide.md
```diff
@@ -398,6 +398,40 @@ Staging binaries
...
+1. In apache-beam-wheels/.travis.yml file:
+   * change RC=current rc
+   * change VERSION=current release version
+   * change WHEELHOUSE_UPLOADER_USERNAM=your test pypi account username 
+   * delete secure line
+1. Generate your own secure
+   * Install ruby dev
+   
+     ```
+     sudo apt-get install ruby2.3-dev
```
jbonofre: It should be `ruby2.5-dev` IMHO.

aaltay:
What is the status of this PR?

boyuanzz:
Need further investigation. Close this PR for now.

PR #474, opened 2018-06-20
Add security vulnerability contact instructions

ThatRfernand:
Adding instructions for reporting a security vulnerability to the project, as discussed in dev list [1]. 

Note that we do not have a security@beam.apache.org mailing list, so the standard process suggests adding security@apache.org as a point of contact.

[1] https://lists.apache.org/thread.html/fd12fd223b65556de5b6e14fa218c329e8d45e98f8212f9658932ef7@%3Cdev.beam.apache.org%3E

### DIFF
--- a/content/community/contact-us/index.html
+++ b/content/community/contact-us/index.html
@@ -154,8 +154,10 @@ whichever one seems best!</p>
     </tr>
   </tbody>
 </table>
-
-
+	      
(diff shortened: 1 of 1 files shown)


aaltay:
@asfgit merge

kennknowles:
This was the generated code. Do the same to the markdown in `src/community/contact-us` or thereabouts.

PR #478, opened 2018-06-26
Change download link for 2.4.0 version to use the archive link

aaltay:
Fix the broken links.

@jbonofre - I could not find the sha-512 and signature files in the archive area, is this intended?


aaltay:
R: @ThatRfernand

aaltay:
@asfgit merge

PR #483, opened 2018-06-28
Move CREATE TABLE and SELECT up a level

apilloud:
This moves CREATE TABLE and SELECT up a level.

@kennknowles


kennknowles:
Note that the file structure doesn't _have_ to match the menu structure.

kennknowles:
From my reading, the tests are spurious. I think the test incorporated with mergebot might be more accurate. I am going to try merging.

kennknowles:
@asfgit merge

PR #485, opened 2018-06-29
Add redirect to new team page

apilloud:
This page is still linked from places (google search for `beam committers`, we should redirect to the new page.


apilloud:
@kennknowles This isn't important, I just wanted to fix it.

kennknowles:
As usual, it cannot be verified because of the way the redirects are generated. I'll test manually.

PR #494, opened 2018-07-12
Minor tweak to Intellij instructions.

tweise:

### DIFF
--- a/src/contribute/intellij.md
+++ b/src/contribute/intellij.md
@@ -30,6 +30,8 @@ limitations under the License.
 
 ## Create a working Gradle module
 
+(as of Intellij 2018.1.6)
+
 1. Create an empty IntelliJ project outside of the Beam source tree.
 2. Under Project Structure > Project, select a Project SDK.
 3. Under Project Structure > Modules, click the + sign to add a module and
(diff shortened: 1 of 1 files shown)


src/contribute/intellij.md
```diff
@@ -41,8 +41,8 @@ limitations under the License.
        * Create separate module per source set
        * Store generated project files externally
        * Use default gradle wrapper
-4. Delegate build actions to Gradle by going to Settings > Build, Execution,
```
iemejia: It seems `Settings/Preferences` is OS specific in linux there is no `Preferences` and since it is Settings in other lines maybe we let `Settings`. Apart of this LGTM.

lukecwik: I believe this is also version dependent. In 2017.1 I believe it was `... > Build Tools > Gradle`

I can confirm that in 2018.1 it is now `... > Build Tools > Gradle > Runner`

How about we compromise on `Preferences/Settings > Build, Execution, Deployment > Build Tools > Gradle > Runner`?

tweise: Yep. Also I'm testing this with 2017.3 - will mention the Intellij version here as well. Do we need to add anything regarding manual classpath tweaks?

tweise:
See update. `Store generated project files externally` is also gone in 2018.1

lukecwik:
@asfgit merge

PR #507, opened 2018-07-20
Fix shell tabs, add new shell tab switcher

melap:
http://apache-beam-website-pull-requests.storage.googleapis.com/507/index.html


melap:
@asfgit merge

PR #565, opened 2018-09-27
[BEAM-4499] Remove non-release documentation sources and generated content

swegner:


swegner:
R: @alanmyrvold

Note that the plan is to merge this into a new `release-docs` branch which will serve only the release-based javadocs / pydocs

swegner:
The diff view here isn't useful; it's easier to see the directory structure from the [tree view](https://github.com/apache/beam-site/tree/8d95bc45cc05533cc6793482ce3dfa8a92ef6432).

swegner:
I believe this is ready for review. @alanmyrvold PTAL?

CC: @udim

README.md
```diff
@@ -1,97 +1,29 @@
...
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+
+This branch contains the generated release documentation for http://beam.apache.org.
+
+To contribute to the  website, please modify the
+[website sources on master](https://github.com/apache/beam/tree/master/website).
```
alanmyrvold: Since this readme is in beam-site, should clarify to modify the website/src in the *beam* repository on the master branch. The link is to beam, but without following the link it is confusing.

.htaccess
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Keep this updated to point to the current release.
+RedirectMatch "/(javadoc|pydoc)/current" "/$1/2.6.0"
```
alanmyrvold: Is there something added to the release guide to update this?

melap: would it be easy to make this redirect subpages also since we're adding this? then folks could use /current/ in links to specific javadoc/pydoc pages.  /current/* -->  /2.6.0/*

if it's not a simple change, I can file a JIRA request for the future. it's P2 for docs on the Beam site due to the site latest version variable that's updated every release, but this is a big pain point for non-Beam docs that point to specific Beam javadoc/pydoc as every link must be updated for each release.

swegner: re: @alanmyrvold on release docs: @udim has a PR out for that; I'll work with him to make sure this gets added.

re: @melap on redirect subpages: yes, actually this is the default behavior! Documentation [here](https://httpd.apache.org/docs/current/mod/mod_alias.html#redirect).

README.md
swegner: Good point. updated.

.htaccess
melap: AWESOME!

PR #571, opened 2018-10-03
Remove htaccess file

swegner:
This configuration didn't actually work. This will be moved into https://github.com/apache/beam/blob/master/website/src/.htaccess instead.

Once your pull request has been opened and assigned a number, please edit the
URL below, replacing `PULL_REQUEST_NUMBER` with the number of your pull request.

http://apache-beam-website-pull-requests.storage.googleapis.com/PULL_REQUEST_NUMBER/index.html

Finally, it will help us expedite review of your Pull Request if you tag
someone (e.g. @username) to look at it.


swegner:
R: @udim

swegner:
The `/current` redirect was added in the main `.htaccess` file in https://github.com/apache/beam/pull/6556 instead. It turns out this `.htaccess` file didn't work because when published the Apache web server it isn't in root directory and thus doesn't get picked up.

PR #580, opened 2018-10-05
Remove extra redirect

swegner:
This is causing issues with pre-commit because the link doesn't exist on beam.apache.org, and the generated redirect files use absolute URLs, rather than relative URLs

**Deprecation notice:** The website is being migrated to
https://github.com/apache/beam/tree/master/website

Please create new pull requests against the above repo.

---

*Please* add a meaningful description for your change here.

Once your pull request has been opened and assigned a number, please edit the
URL below, replacing `PULL_REQUEST_NUMBER` with the number of your pull request.

http://apache-beam-website-pull-requests.storage.googleapis.com/PULL_REQUEST_NUMBER/index.html

Finally, it will help us expedite review of your Pull Request if you tag
someone (e.g. @username) to look at it.

### DIFF
--- a/src/documentation/dsls/sql/create-external-table.md
+++ b/src/documentation/dsls/sql/create-external-table.md
@@ -5,7 +5,6 @@ section_menu: section-menu/sdks.html
 permalink: /documentation/dsls/sql/create-external-table/
 redirect_from:
   - /documentation/dsls/sql/statements/create-table/
-  - /documentation/dsls/sql/create-table/
 ---
 <!--
 Licensed under the Apache License, Version 2.0 (the "License");


swegner:
R: @alanmyrvold @udim

swegner:
yeah, with more thought I guess this will always be broken when you use `redirect_from`:

The scenario for using a redirect is if you're moving `pageA.md` to `pageB.md`; you would rename the file to `pageB.md` and add `redirect_from: pageA`. This will cause jekyll to generate a `pageA.html` with a redirect to `https://beam.apache.org/pageB.html`. But `pageB.html` doesn't exist on the livesite yet, so it will always fail.

PR #582, opened 2018-10-18
Remove 2.7.0 pydoc .doctrees directory

swegner:
According to the release guide, this should not be checked in:
https://beam.apache.org/contribute/release-guide


swegner:
R: @aaltay 
CC: @udim

udim (approved):

PR #585, opened 2019-01-28
Update beam-site javadoc and pydoc for release 2.10.0

kennknowles:


kennknowles:
Will need to roll RC2. In case it changes the javadoc or pydoc I will just regenerate.

PR #626, opened 2022-04-25
`current` symlinks should use relative paths

TheNeuralBit:
Context: https://lists.apache.org/thread/158mg25sfdgltzsjx7sr5ks4q6zhkyhq
See also: https://github.com/apache/beam/pull/17457


TheNeuralBit:
R: @damccorm

damccorm (approved):

PR #673, opened 2024-11-15
Stage pydoc/javadoc/etc..

damccorm:
This follows https://github.com/apache/infrastructure-asfyaml?tab=readme-ov-file#web-site-deployment-service-for-git-repositories to add a staging site. I think this is the right way to do it, but I'm not totally sure. Worst case, we can revert.

I want to do this so that we can test https://github.com/apache/beam/issues/32901

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -16,6 +16,10 @@
 
 # This file is required to publish the Beam Docs.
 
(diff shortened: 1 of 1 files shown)


damccorm:
R: @liferoad

liferoad (approved):

